PR #106, opened 2023-10-12
User Agent, non-news domains, batch logging

pgulley:
Reopened without batching concenrs

### DIFF
--- a/indexer/workers/fetcher/fetch_worker.py
+++ b/indexer/workers/fetcher/fetch_worker.py
@@ -7,6 +7,7 @@ from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, TypedDict
 
+from mcmetadata.urls import NON_NEWS_DOMAINS
 from scrapy.crawler import CrawlerProcess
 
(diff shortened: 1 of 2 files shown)


philbudne (approved):

indexer/workers/fetcher/fetch_worker.py
```diff
@@ -134,16 +135,29 @@ def scrapy_cb(self, story: BaseStory) -> None:
         else:
             status_label = f"http-{http_meta.response_code//100}xx"
 
+        assert http_meta.final_url is not None
+        if any(dom in http_meta.final_url for dom in NON_NEWS_DOMAINS):
+            status_label = "non-news"
+
+        else:
+            self.fetched_stories.append(story)
```
kilemensi, line 143: It's been a while since I looked at fetcher, so my understanding could be wrong: shouldn't we just check if a `final_url` is in `NON_NEWS_DOMAINS` when `response_code == 200`? Do we care if `final_url` is in `NON_NEWS_DOMAINS` but the `response_code == 404` for example? It not going to get into ElasticSearch any way.

PR #109, opened 2023-10-12
Es indexing update

thepsalmist:
This PR introduces;
1 .Set index namesmediacloud_search_text_older for pre 2020 dates and mediacloud_search_text_other for NULL dates and future publication dates
2. Remove ELASTICSEARCH_INDEX_NAME, not required param at initialization

### DIFF
--- a/indexer/tests/test_importer.py
+++ b/indexer/tests/test_importer.py
@@ -22,7 +22,7 @@ def set_env() -> None:
     os.environ["ELASTICSEARCH_HOSTS"] = ",".join(
         ["http://localhost:9200", "http://localhost:9201", "http://localhost:9202"]
     )
-    os.environ["ELASTICSEARCH_INDEX_NAME"] = "test_mediacloud_search_text_older"
+    # os.environ["ELASTICSEARCH_INDEX_NAME"] = "test_mediacloud_search_text_older"
     os.environ["ELASTICSEARCH_INDEX_NAME_PREFIX"] = "test_mediacloud_search_text"
 
 
@@ -38,14 +38,14 @@ def elasticsearch_client() -> Any:
     assert hosts is not None, "ELASTICSEARCH_HOSTS is not set"
     client = create_elasticsearch_client(hosts=hosts)
     assert client.ping(), "Failed to connect to Elasticsearch"
-    index_name = os.environ.get("ELASTICSEARCH_INDEX_NAME")
-    assert index_name is not None, "ELASTICSEARCH_INDEX_NAME is not set"
+    index_name_prefix = os.environ.get("ELASTICSEARCH_INDEX_NAME_PREFIX")
+    assert index_name_prefix is not None, "ELASTICSEARCH_INDEX_NAME_PREFIX is not set"
 
-    recreate_indices(client, index_name)
+    recreate_indices(client, index_name_prefix)
 
     yield client
 
-    recreate_indices(client, index_name)
+    recreate_indices(client, index_name_prefix)
 
 
 test_data: Mapping[str, Optional[Union[str, bool]]] = {
@@ -91,11 +91,7 @@ class TestElasticsearchConnection:
 @pytest.fixture(scope="class")
 def elasticsearch_connector() -> ElasticsearchConnector:
     elasticsearch_hosts = cast(str, os.environ.get("ELASTICSEARCH_HOSTS"))
-    index_name = os.environ.get("ELASTICSEARCH_INDEX_NAME")
-    assert index_name is not None, "ELASTICSEARCH_INDEX_NAME is not set"
-    connector = ElasticsearchConnector(
-        elasticsearch_hosts, index_name, es_mappings, es_settings
-    )
+    connector = ElasticsearchConnector(elasticsearch_hosts, es_mappings, es_settings)
     return connector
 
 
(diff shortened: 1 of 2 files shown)


indexer/workers/importer.py
```diff
@@ -92,8 +88,12 @@ def create_index(self, index_name: str) -> None:
                     mappings=self.mappings,
                     settings=self.settings,
                 )
+                logger.info(f"Index '{index_name}' created successfully.")
             else:
                 self.client.indices.create(index=index_name)
+                logger.info(f"Index '{index_name}' created successfully.")
+        else:
+            logger.warning(f"Index '{index_name}' already exists. Skipping creation.")
```
kilemensi, line 96: If I understand the code flow correctly, every time we want to import a document, we call `index`, which in turn calls `create_index` to check if the index does indeed exist in ES, and logs a _**warning**_ if the index already exists.

That'll be a lot of warnings in the log files!... probably defeating the purpose of a warning.

indexer/workers/importer.py
```diff
@@ -154,14 +153,23 @@ def index_routing(self, publication_date_str: Optional[str]) -> str:
         if publication_date_str:
             try:
                 year = datetime.strptime(publication_date_str, "%Y-%m-%d").year
+                current_year = datetime.now().year
+
+                if year > current_year:
+                    year = -1
+                    logger.warning(
+                        f"Publication date greater than current year: {current_year}"
+                    )
```
kilemensi, line 162: 1. The if year > current_year check can be done outside try/except since it's not going to raise.
2. There was a talk about 90 days in the future... How does that play into this (if it does, not too sure)?

indexer/workers/importer.py
```diff
@@ -154,14 +153,23 @@ def index_routing(self, publication_date_str: Optional[str]) -> str:
...
             except ValueError as e:
-                logger.error(f"Error parsing date: {str(e)}")
+                logger.warning(f"Error parsing date: {str(e)}")
 
         index_name_prefix = os.environ.get("ELASTICSEARCH_INDEX_NAME_PREFIX")
-        if year and year >= 2021:
+        if 2021 <= year <= current_year:
```
kilemensi, line 167:
```suggestion
        if year >= 2021:
```
Since you've already handled the `if year > current_year` part.

indexer/tests/test_importer.py
```diff
@@ -128,6 +123,6 @@ def test_index_routing(
         index_name_prefix = os.environ.get("ELASTICSEARCH_INDEX_NAME_PREFIX")
         assert index_name_prefix is not None
         assert importer.index_routing("2023-06-27") == f"{index_name_prefix}_2023"
-        assert importer.index_routing(None) == f"{index_name_prefix}_older"
+        assert importer.index_routing(None) == f"{index_name_prefix}_other"
```
kilemensi, line 126: May be one test for future date going into `_other` as well?

indexer/workers/importer.py
philbudne: I agree, I think it might be better as a debug message, and to avoid the overhead of formatting a message that may not be displayed, I recommend passing a percent-format and variables to logger.debug and logger.info calls (or all logger message methods for that matter)

indexer/workers/importer.py
philbudne: ISTR the 90 day future check is in mcmetadata, maybe this: https://github.com/mediacloud/metadata-lib/blob/72c44939e57ea208063f56fcb7427f0bd6f83c08/mcmetadata/__init__.py#L18

indexer/workers/importer.py
```diff
@@ -92,8 +88,12 @@ def create_index(self, index_name: str) -> None:
                     mappings=self.mappings,
                     settings=self.settings,
                 )
+                logger.info("Index '%s' created successfully." % index_name)
             else:
                 self.client.indices.create(index=index_name)
+                logger.info("Index '%s' created successfully." % index_name)
```
kilemensi, line 94:
```suggestion
            else:
                self.client.indices.create(index=index_name)
            logger.info("Index '%s' created successfully." % index_name)
```
This should log for either `if` or `else` path I believe.

indexer/workers/importer.py
```diff
@@ -153,15 +152,21 @@ def index_routing(self, publication_date_str: Optional[str]) -> str:
         year = -1
         if publication_date_str:
             try:
-                year = datetime.strptime(publication_date_str, "%Y-%m-%d").year
+                pub_date = datetime.strptime(publication_date_str, "%Y-%m-%d")
+                year = pub_date.year
+                # check for exceptions of future dates just in case gets past mcmetadata
+                if pub_date > datetime.now() + timedelta(days=+90):
```
kilemensi, line 158:
```suggestion
                if pub_date > datetime.now() + timedelta(days=90):
```
Why the `+`?

... and that `90` should probably be a constant declared somewhere.

indexer/workers/importer.py
```diff
@@ -153,15 +152,21 @@ def index_routing(self, publication_date_str: Optional[str]) -> str:
...
+                    year = -1
             except ValueError as e:
-                logger.error(f"Error parsing date: {str(e)}")
+                logger.warning("Error parsing date: '%s" % str(e))
 
         index_name_prefix = os.environ.get("ELASTICSEARCH_INDEX_NAME_PREFIX")
```
kilemensi, line 163:
```suggestion
```
I think you should just use `self.index_name_prefix` which is already initialised in the `process_args` method.

kilemensi (approved):

philbudne (approved):

Issue #103, opened 2023-10-09
elasticsearch index stats in grafana


philbudne:
@rahulbot would it be useful to have stats for elasticsearch indices in grafana?

line breaks inserted below for clarity:
```
>>> r = requests.get('http://elasticsearch1:9200/_cat/indices?format=json')
>>> r.text
'[{"health":"green","status":"open","index":"mediacloud_search_text_v2","uuid":"B3HVWUMNRYa8cdej_4m5Yw","pri":"1","rep":"1","docs.count":"1715302","docs.deleted":"0","store.size":"13.2gb","pri.store.size":"6.6gb"},
{"health":"green","status":"open","index":"mediacloud_search_text_2023","uuid":"_OVWbmIORfGn-H26PUfI4Q","pri":"1","rep":"0","docs.count":"242004","docs.deleted":"0","store.size":"1.7gb","pri.store.size":"1.7gb"},
{"health":"green","status":"open","index":"mediacloud_search_text_2022","uuid":"8buLDxisQoCo_hPSTZdvvQ","pri":"1","rep":"0","docs.count":"1079","docs.deleted":"0","store.size":"5.5mb","pri.store.size":"5.5mb"},
{"health":"green","status":"open","index":"mediacloud_search_text_2021","uuid":"_3cg_4-TQIC2oFsMGJDlag","pri":"1","rep":"0","docs.count":"1068","docs.deleted":"0","store.size":"5.2mb","pri.store.size":"5.2mb"},
{"health":"green","status":"open","index":"is_this_still_needed","uuid":"o1UxSRjBSIyhIbuC0FhxSg","pri":"1","rep":"0","docs.count":"0","docs.deleted":"0","store.size":"247b","pri.store.size":"247b"},
{"health":"green","status":"open","index":"mediacloud_search_text_older","uuid":"XXQuLYVLSi6o52H39YmVMw","pri":"1","rep":"0","docs.count":"34924","docs.deleted":"0","store.size":"338.5mb","pri.store.size":"338.5mb"},
{"health":"green","status":"open","index":"mediacloud_search_text","uuid":"vh0Q3pJjSPKHdbQLgCB1uw","pri":"1","rep":"1","docs.count":"2544074","docs.deleted":"2","store.size":"20.1gb","pri.store.size":"10gb"}]'
```

philbudne:
PR has been merged.

PR #110, opened 2023-10-12
Report Elastic Search stats to graphite-grafana

philbudne:
*Low priority*, for https://github.com/mediacloud/story-indexer/issues/103

* Creates indexer/scripts/elastic-stats.py. bin/run-elastic.stats and an elastic-stats container
* Creates indexer.app.ArgsProtocol for App mixins
* Creates indexer.app.IntervalMixin with `--interval` option for use by {elastic,rabbitmq}-stats.py
* Creates indexer.elastic.ElasticMixin with `--elasticsearch-hosts` with code from workers/importer.py: use in importer and elastic-stats

This is my first attempt at writing mixin classes in a project checked by mypy, so I could well be doing it wrong!

I abstracted code from importer.py and rabbitmq-stats.py to create two mixins for common command-line arguments, and their implementation.

Because the `ElasticMixin.elasticsearch_client` method returns an `Elasticsearch` client object I modified the indexer `ElasticsearchConnector` to take a client object rather than the fully general possibilities that the `Elasticsearch` constructor allows, and it seems like it does the right thing a list of URLs, but I'm wondering if I'm missing a reason for the old way (converting the URLs to `NodeConfig` objects)??

elastic-stats reports `docs` (document count), `deleted` (deleted/duplicate documents) and `pri-size` (size of primary index in MB) for each index, as well as counts of indices in each color/state: green/yellow/red, which could be useful for alerting.

I chose to use the short name "elastic" in filenames rather than the longer "elasticsearch", but I could be convinced otherwise!!

### DIFF
--- a/indexer/elastic.py
+++ b/indexer/elastic.py
@@ -16,7 +16,7 @@ from elasticsearch import Elasticsearch
 
 from indexer.app import ArgsProtocol
 
-logger = getLogger("elastic-stats")
+logger = getLogger(__name__)
 
 
 class ElasticMixin:
--- a/indexer/elastic.py
+++ b/indexer/elastic.py
@@ -14,28 +14,30 @@ from urllib.parse import urlparse
 from elastic_transport import NodeConfig, ObjectApiResponse
 from elasticsearch import Elasticsearch
 
+from indexer.app import ArgsProtocol
+
 logger = getLogger("elastic-stats")
 
-# would prefer to have an ElasticMixin for Apps, but would need a
-# "protocol" to inherit from, so a bunch of functions to call
-# from App methods.
 
+class ElasticMixin:
+    """
+    mixin class for Apps that use Elastic Search API
+    """
 
-def add_elasticsearch_hosts(ap: argparse.ArgumentParser) -> None:
-    ap.add_argument(
-        "--elasticsearch-hosts",
-        dest="elasticsearch_hosts",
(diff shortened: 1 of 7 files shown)


indexer/elastic.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def elasticsearch_client(self: ArgsProtocol) -> Elasticsearch:
+        assert self.args
+        hosts = self.args.elasticsearch_hosts
+        if not hosts:
+            logger.fatal("need --elasticsearch-hosts or ELASTICSEARCH_HOSTS")
+            sys.exit(1)
+
+        # Connects immediately, performs failover and retries
+        return Elasticsearch(hosts.split(","))
```
thepsalmist, line 45: The issue I faced with just passing the hosts to create the ES client like this was this `mypy` error 
`error: Argument 1 to "Elasticsearch" has incompatible type "list[str]"; expected "str | list[str | Mapping[str, str | int] | NodeConfig] | None"  [arg-type]`

philbudne: I ran into the same issue, but mypy doesn't have complaints with this version of the code!

pgulley (approved):

kilemensi (approved):

philbudne:
@thepsalmist let me know if I can merge this (I know you've had time off), or if I need to make changes in your opinion.

thepsalmist (approved):

Issue #112, opened 2023-10-18
Put bare metal ES server configs in repo?


philbudne:
In the interests of documentation and recoverability, should we put bare metal configs in this repo
(so long as they don't contain sensitive info)?

philbudne:
There are other config items we'll want to store:

Configuring node labels for use by docker-compose file

rabbitmq configs (if/when we get there)

I know in the past MC has used ansible for configuration, and for multiple hosts, it _might_ make sense.
My past experience with ansible is that it takes me 3x longer to do anything than if I wrote a shell script!

rahulbot:
Agreed. Can this be changed into a shell script that would automate the majority of the setup of a clean machine based on some env-vars?

PR #113, opened 2023-10-18
Update ES configuration script and documentation

thepsalmist:
This PR introduces
1. Updating ES configuration documentation from the bare metal config.
2. Introducing an ES deploy script that can be used to configure new ES instances on bare metal

Handles issues: #112 , #142


philbudne (approved):

doc/elasticsearch/elasticsearch.yml
```diff
@@ -0,0 +1,125 @@
...
+# address here to expose this node on the network:
+#
+network.host: _eno1_
+#
+# By default Elasticsearch listens for HTTP traffic on the first free port it
+# finds starting at 9200. Set a specific HTTP port here:
+
+# PS:Current configuration based on available ports, not used by docker containers
+http.port: 9200
+transport.port: 9305
```
kilemensi, line 63: We no longer need to worry about docker, right?

doc/elasticsearch/elasticsearch.yml
```diff
@@ -0,0 +1,125 @@
...
+xpack.security.http.ssl:
+  enabled: false
+  keystore.path: certs/http.p12
+
+# Enable encryption and mutual authentication between cluster nodes
+xpack.security.transport.ssl:
+  enabled: false
+  verification_mode: certificate
+  keystore.path: certs/transport.p12
+  truststore.path: certs/transport.p12
```
kilemensi, line 112: Looks like we're **disabling** security so may be just one line comment to say that and why (probably because ES is not exposed to the Internet).

doc/elasticsearch/elasticsearch_configuration.md
```diff
@@ -0,0 +1,116 @@
...
+
+`
+wget https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-8.9.2-amd64.deb`
+
+`
+wget https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-8.9.2-amd64.deb.sha512`
+
+`shasum -a 512 -c elasticsearch-8.9.2-amd64.deb.sha512 `
+
+`sudo dpkg -i elasticsearch-8.9.2-amd64.deb`
```
kilemensi, line 15: Isn't it better to install from APT [repository](https://www.elastic.co/guide/en/elasticsearch/reference/current/deb.html#deb-key)? That way we may benefit from security updates.

pgulley (approved):

doc/elasticsearch/elasticsearch.yml
philbudne: This config is for production, so it never runs under docker, and should be using the default port numbers in all cases (and if not, production should be readjusted to use the default port, and config files with the minimum number of changes needed (so it's easier to update to a new release)

The staging stack runs ES under docker on the same cluster of servers, but should be using default port + 10 in all cases via deploy.sh and docker-compose.yml.j2 parameters (and if not, staging config should be adjusted, not production)

PR #114, opened 2023-10-18
ES snapshots creation

thepsalmist:
This PR introduces a worker that would take regular ES snapshots and backup to S3.
The S3 repository setup is done at configuration, and the registered repository name passed as an arg `ELASTICSEARCH_SNAPSHOT_REPO`


philbudne (commented):
Whoops!  Made comments, earlier, but they were stuck in "pending" state!

indexer/elastic.py
```diff
@@ -34,6 +34,13 @@ def define_options(self: ArgsProtocol, ap: argparse.ArgumentParser) -> None:
             help="comma separated list of ES server URLs",
         )
 
+        ap.add_argument(
```
philbudne, line 37: Looks good!!

Moving `--elasticsearch-snapshot-repo` here means it will appear as an option to elastic-stats.py

If you anticipate other programs will need the option, then I suggest making an `ElasticSnapshotMixin` subclass of `ElasticMixin`, otherwise I think it should stay in elastic-snapshots.py

pgulley (approved):

indexer/scripts/elastic-snapshots.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+        snapshot_info = response.get("snapshot", {})
+        if snapshot_info.get("state") == "SUCCESS":
+            logger.info(f"Successfully created {snapshot_name} in S3")
+            logger.info(f"Start Time: {snapshot_info.get('start_time')}")
+            logger.info(f"End Time: {snapshot_info.get('end_time')}")
+            shards_info = snapshot_info.get("shards", {})
+            logger.info(
+                f"Shards - Total: {shards_info.get('total', 0)}, Failed: {shards_info.get('failed', 0)}, Successful: {shards_info.get('successful', 0)}"
+            )
```
kilemensi, line 46: Works but I think a single line log if possible is better (storage and searching) e.g. `snapshot xxx created T:start_time/end_time S:#success/#failure/#total` where `T` is time and `S` is shard and for shard.

Or if we know the endpoint, instead of `snapshot xxx created ...`, you could have `ACTION /endpoint 200 ...` similar to how other HTTP requests are logged.

kilemensi (commented):
👍🏽 

---


What Phil said and may be if we fail to create a snapshot it should be _critical_ rather than just _error_. I'm assuming this is a kind of a thing we'd want to know immediately.

indexer/elastic.py
philbudne: Again, I don't think we need/want this option to appear for the importer.

thepsalmist: Resolved

indexer/scripts/elastic-snapshots.py
thepsalmist: Didn't exactly get option 2, but restructure the logged info, and added counters for "success" and "failures"

indexer/elastic.py
```diff
@@ -43,3 +43,24 @@ def elasticsearch_client(self: ArgsProtocol) -> Elasticsearch:
...
+        super().define_options(ap)
+
+        ap.add_argument(
+            "--elasticsearch-snapshot-repo",
+            dest="elasticsearch_snapshot_repo",
+            default=os.environ.get("ELASTICSEARCH_SNAPSHOT_REPO") or "",
+            help="ES snapshot repository name",
+        )
+
+    def get_elasticsearch_snapshot(self: ArgsProtocol) -> Any:
```
kilemensi, line 59: Isn't the repo standard to have a `process_args` method that will ensure all required args have been set on `self`?

indexer/scripts/elastic-snapshots.py
```diff
@@ -0,0 +1,57 @@
...
+    def main_loop(self) -> None:
+        _TODAY = date.today().strftime("%Y.%m.%d")
+        repository_name: str = self.get_elasticsearch_snapshot()
+        snapshot_name: str = f"snapshot-{_TODAY}"
+
+        client = self.elasticsearch_client()
+        assert client.ping(), "Failed to connect to Elasticsearch"
+
+        index_names = list(client.indices.get_alias().keys())
+        indices: str = ",".join(index_names)
```
kilemensi, line 25: mypy can't infer this?

indexer/scripts/elastic-snapshots.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+        snapshot_info = response.get("snapshot", {})
+        if snapshot_info.get("state") == "SUCCESS":
+            self.incr("snapshots", labels=[("status", "success")])
+            shards_info = snapshot_info.get("shards", {})
+            logger.info(
+                f"Snapshot Created Successfully Information - "
+                f"N: {snapshot_name} "
+                f"T: {snapshot_info.get('start_time')}/{snapshot_info.get('end_time')} "
+                f"S: {shards_info.get('total', 0)}/{shards_info.get('failed', 0)}"
```
kilemensi, line 44: I think we're using `%` formatting for log messages.

philbudne (approved):
LGTM

indexer/scripts/elastic-snapshots.py
```diff
@@ -0,0 +1,77 @@
...
+                "N: %s "
+                "T: %s/%s "
+                "S: %s/%s"
+                % (
+                    snapshot_name,
+                    snapshot_info.get("start_time"),
+                    snapshot_info.get("end_time"),
+                    shards_info.get("total", 0),
+                    shards_info.get("failed", 0),
+                )
```
kilemensi, line 64:
```suggestion
                "S: %s/%s",
                snapshot_name,
                snapshot_info.get("start_time"),
                snapshot_info.get("end_time"),
                shards_info.get("total", 0),
                shards_info.get("failed", 0),
                )
```

params should be passed into `logger.info`... otherwise, we're doing the string formatting.

kilemensi (approved):

Issue #118, opened 2023-10-19
assertion failures seen in fetcher


philbudne:
While examining log files during testing I saw:
```
2023-10-19 04:17:22,644 | INFO | scrapy.extensions.throttle | slot: noticias.uol.com.br | conc: 1 | delay: 5000 ms (+0) | latency:  303 ms | size:   456 bytes
2023-10-19 04:17:22,645 | DEBUG | scrapy.core.engine | Crawled (403) <GET https://noticias.uol.com.br/politica/ultimas-noticias/2023/10/17/relatorio-da-cpi-de-81-pede-indiciamento-de-bolsonaro-por-tentar-golpe.htm> (referer: None)
2023-10-19 04:17:22,747 | ERROR | scrapy.core.scraper | Spider error processing <GET https://noticias.uol.com.br/politica/ultimas-noticias/2023/10/17/relatorio-da-cpi-de-81-pede-indiciamento-de-bolsonaro-por-tentar-golpe.htm> (referer: None)
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/twisted/internet/defer.py", line 892, in _runCallbacks
    current.result = callback(  # type: ignore[misc]
  File "/app/indexer/workers/fetcher/batch_spider.py", line 81, in on_error
    self.cb(story)
  File "/app/indexer/workers/fetcher/fetch_worker.py", line 138, in scrapy_cb
    assert http_meta.final_url is not None
AssertionError
```

Perhaps both assertions in indexer/workers/fetcher/batch_spider.py should be if statements that set `status_label` to make the events visible?

philbudne:
Is there a counter for "failed to make a connection"?  rss-fetcher splits this out (in perhaps too many counters).
A counter for "too many redirects"?

philbudne:
I'm pretty sure the assert failures are a thing of the past, BUT I'd still be interested in seeing counters for "connection failed" and "too many redirects"; so perhaps this bug should be re-titled?

rahulbot:
Task: see if this is still happening

philbudne:
No evidence that the problem still exists (checked logs from all 20 batches of last fetch).
Closing.

PR #129, opened 2023-10-23
under/oversized stories counters

pgulley:
Straightforward

### DIFF
--- a/indexer/workers/fetcher/fetch_worker.py
+++ b/indexer/workers/fetcher/fetch_worker.py
@@ -182,8 +182,6 @@ class FetchWorker(QApp):
         assert self.connection
         chan = self.connection.channel()
 
-        queued_stories = 0
-        oversized_stories = 0
         for story in self.fetched_stories:
             http_meta = story.http_metadata()
 
@@ -196,18 +194,9 @@ class FetchWorker(QApp):
                     logger.warn(
                         f"Story over {MAX_FETCHER_MSG_SIZE} limit: {story.rss_entry().link}, size: {len(story_dump)}"
                     )
-                    oversized_stories += 1
+                    self.incr("oversized-stories")
                 else:
                     self.send_message(chan, story_dump)
-                    queued_stories += 1
-
-        self.gauge(
-            "oversized-stories", oversized_stories, labels=[("batch", self.batch_index)]
-        )
-
-        self.gauge(
-            "queued-stories", queued_stories, labels=[("batch", self.batch_index)]
-        )
 
 
 if __name__ == "__main__":
--- a/indexer/workers/fetcher/fetch_worker.py
+++ b/indexer/workers/fetcher/fetch_worker.py
@@ -134,8 +134,16 @@ class FetchWorker(QApp):
         else:
             status_label = f"http-{http_meta.response_code//100}xx"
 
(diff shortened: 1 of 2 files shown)


indexer/workers/parser.py
```diff
@@ -40,8 +40,10 @@ def process_story(
         #     or let fail from repeated retries???
         try:
             mdd = mcmetadata.extract(link, html)
-        except mcmetadata.exceptions.BadContentError as e:
-            raise QuarantineException(getattr(e, "message", repr(e)))
+        except mcmetadata.exceptions.BadContentError:
+            self.incr(
+                "undersized-stories"
+            )  # _PG: Or "unparsed-story" label ('reason', 'too-short'), and add counters for no html and no link too?
```
philbudne, line 46: A counter with multiple labels is easier to deal with (with thing.* queries) than multiple counter names (so long as only one label gets incremented).

I'd vote for a simple "stories" counter with status=ok, (or if you did status=ok-METHOD you could remove the parsed-stories counter)

indexer/workers/fetcher/fetch_worker.py
```diff
@@ -196,18 +194,9 @@ def main_loop(self) -> None:
                     logger.warn(
                         f"Story over {MAX_FETCHER_MSG_SIZE} limit: {story.rss_entry().link}, size: {len(story_dump)}"
                     )
-                    oversized_stories += 1
+                    self.incr("oversized-stories")
```
philbudne, line 197: As below (wrote that comment first), a single "stories" counter with different status values would mean that when a new status is added, it appears in the graphs without human intervention.

If the test is done based on HTML size alone (ie; `FETCHER_MAX_HTML`; which is a reasonable proxy, and easier for muggles to understand), the test can be done in `scrapy_cb` (along with replacing the various asserts there with different status values).  Logging the URL is good: it means we can grep the logs to see what sources are responsible.

philbudne: regarding asserts: https://github.com/mediacloud/story-indexer/issues/118 for log messages from an assertion failure

pgulley: Good idea to move the check into scrapy_cb! With a single counter in the fetcher, I'm concerned that a "status=oversized" label would overwrite the "http_xxx" label for oversized stories. I suppose that we don't anticipate very many oversized stories ever, so something like "status=oversized_http_xxx" might do?

philbudne:
> "status=oversized_http_xxx"

Hmm.  I thought the only HTTP status that returned valid data was 200!

The rss-fetcher doesn't accept anything other than 200:
https://github.com/mediacloud/rss-fetcher/blob/main/fetcher/tasks.py#L830
(304 happens if we've set conditions on when NOT to send an unchanged document)

Is there a way to tell an HTTP server we won't accept bodies above a
certain size that scrapy uses??

Even if so, I'd say http_xxx and oversized are fine as is; there will
always be different causes for rejection for similar reasons, and
trying to interpret and align data at the collection level will be, at
best incomplete, and possibly confusing.

I try to keep counter names short, so they don't get truncated when
viewed.

philbudne (approved):

philbudne:
@pgulley is this ready for merge?

PR #136, opened 2023-10-27
DRAFT: archiver (now writes WARC files, S3 code not yet tested).

philbudne:
I had ripped the first two implementations of batching out when I added the Pika thread.


pgulley (commented):
This feels like a good start- the actual archiving work will be easy to slip in here when we have agreement on how to do that.

kilemensi (approved):

Issue #135, opened 2023-10-25
add "downloaded date" to index?


rahulbot:
We might want to store and surface the original date of content download - when the HTML data was actually downloaded from the URL in our system.

I've asked the research team and will loop back with a decision.

rahulbot:
The research term agreed that we do _not_ need to store this in the index. However, as indicated in #24 we might still want to pass the `downloaded_date` all the way through so that we can store that metadata in the archive file.

PR #137, opened 2023-10-27
Update ES mapping to index story's first captured time

### LINKED ISSUE #90: importer error handling
Error/exception handling in the indexer predates the addition of retries and quarantine, and likely needs to be re-examined (it's currently catching "Exception", and reports two different conditions w/ status=failed).

Choices for handling an error/exception:

If the error is well-understood, and no recourse is possible, and there is no value in retaining the Story for additional examination and diagnosis, then there should be a status string (counter) for that condition, and the Story can be discarded.

If no recourse is possible, but there is value in retaining the Story for additional examination, a QuarantineException should be thrown with a useful description, either a fixed string or repr(exception).

For a database connection error: close (or abandon) the connection in such a way that the next call to process_story will attempt to re-open the connection, and re-raise the exception, so that the Story will be retried.

Transient conditions that require no special handling need not be caught, see below.

Otherwise unknown conditions should NOT be caught (don't catch Exception or RuntimeError); the story will be retried a finite number of times, and then quarantined.

### LINKED ISSUE #130: add "indexed time" field in ES index
Many regular consumers of query results want to be able to filter for stories matching a query since they last checked for it. The feminicides project is an example -- where we want to list stories that have been captured since a date it tracks internally. We can't use publication date for that need, because discovery date is only loosely correlated with it.

It looks like original integration with IA's index had a [`first_captured`](https://github.com/mediacloud/news-search-api/blob/f91898e49e204e5afeefb4f9b0fc6cbbd7c78088/api.py#L215C19-L215C33) field that held this date time information.  The [current `es_mappings`](https://github.com/mediacloud/story-indexer/blob/6cbacf719fd8d5428e498e4344a97f054be8cfe8/indexer/workers/importer.py#L27C16-L27C16) don't seem to have a similar column.

This is leading to the staging news-search-dashboard current showing an "archived" date for stories on 1970-01-01:
Media_Cloud_—_Staging_Search_API_Explorer

We need to add this field to the ES index to the use-case described above, including other reasons. I think this means (1) adding the field to the mappings, (2) adding the field to the existing indexes, and (3) changing [`ElasticsearchImporter.process_story`](https://github.com/mediacloud/story-indexer/blob/main/indexer/workers/importer.py#L138) to set the property to the current time when it tries to run the import. It'd be fine is all the existing data gets set to "now" and accurate data is only set going forward on new stories.

### LINKED ISSUE #141: Keeping duplicates out of archive
Since the ES index `_id` we generate is a hash of the (normalized?) URL, we should not insert duplicates into ES (so long as the published date remains the same, since different years end up in different indices).

The importer is currently using the `index` operation, which always overwrites existing entries (using `PUT /<target>/_doc/<_id>`), it looks like the response `result` field may indicate whether the entry was `created` or `updated` which could be passed (along with the generated _id?) as Story importer_metadata for use by the archiver.

*HOWEVER*

1. This means the ES index will contain the most recent version of the page, while the archive will get the original version only.
2. When (not if) the RSS fetcher generates a duplicate URL (ie; for a feed which is static and the rss-fetcher prunes its stories table by age) we will continue to write to older indices possibly causing continuing backups of the older indices.

The elasticsearch library `create` operation uses `PUT /<target>/_create/<_id>` which refuses to overwrite existing entries (as does the `index` call if passed `op_type="create"` which would mean we would write only the first instance of a URL/story, and the Story could be dropped, or marked as duplicate.

In any case, the importer should have different counters for new vs duplicate stories!!

thepsalmist:
This PR introduces 
1. Adding a story's `indexed_date` to the ES mapping.
2. Using ES's `index API` creates/updates a document - to ensure duplicates are not indexed, we replace the indexing functionality to use `create API`
3. On indexing, we may get `BadRequest` errors from parsing documents during indexing, this stories are Quarantined to be monitored via `400-*` status counter. Duplicate stories also monitored via incremental counter

Fixes #130 
Fixes #141
Fixes #90

### DIFF
--- a/indexer/elastic.py
+++ b/indexer/elastic.py
@@ -19,25 +19,6 @@ from indexer.app import ArgsProtocol
 logger = getLogger(__name__)
 
 
-def create_elasticsearch_client(
-    hosts: Union[str, List[Union[str, Mapping[str, Union[str, int]], NodeConfig]]],
-) -> Elasticsearch:
-    if isinstance(hosts, str):
-        host_urls = hosts.split(",")
-
-    host_configs: Any = []
-    for host_url in host_urls:
-        parsed_url = urlparse(host_url)
-        host = parsed_url.hostname
-        scheme = parsed_url.scheme
-        port = parsed_url.port
-        if host and scheme and port:
-            node_config = NodeConfig(scheme=scheme, host=host, port=port)
-            host_configs.append(node_config)
-
-    return Elasticsearch(host_configs)
-
-
 class ElasticMixin:
     """
     mixin class for Apps that use Elastic Search API
@@ -60,6 +41,4 @@ class ElasticMixin:
             logger.fatal("need --elasticsearch-hosts or ELASTICSEARCH_HOSTS")
             sys.exit(1)
 
-        # Connects immediately, performs failover and retries
-        es_client = create_elasticsearch_client(hosts)
-        return es_client
+        return Elasticsearch(hosts.split(","))
--- a/indexer/elastic.py
+++ b/indexer/elastic.py
@@ -8,7 +8,7 @@ import argparse
 import os
 import sys
 from logging import getLogger
-from typing import Any, List, Optional
+from typing import Any, List, Mapping, Optional, Union
 from urllib.parse import urlparse
 
 from elastic_transport import NodeConfig, ObjectApiResponse
@@ -17,6 +17,25 @@ from elasticsearch import Elasticsearch
 from indexer.app import ArgsProtocol
 
 logger = getLogger(__name__)
+
+
+def create_elasticsearch_client(
+    hosts: Union[str, List[Union[str, Mapping[str, Union[str, int]], NodeConfig]]],
+) -> Elasticsearch:
+    if isinstance(hosts, str):
+        host_urls = hosts.split(",")
+
+    host_configs: Any = []
+    for host_url in host_urls:
+        parsed_url = urlparse(host_url)
+        host = parsed_url.hostname
+        scheme = parsed_url.scheme
+        port = parsed_url.port
+        if host and scheme and port:
+            node_config = NodeConfig(scheme=scheme, host=host, port=port)
+            host_configs.append(node_config)
+
+    return Elasticsearch(host_configs)
 
 
 class ElasticMixin:
@@ -42,4 +61,5 @@ class ElasticMixin:
             sys.exit(1)
 
         # Connects immediately, performs failover and retries
-        return Elasticsearch(hosts.split(","))
+        es_client = create_elasticsearch_client(hosts)
+        return es_client
--- a/indexer/elastic.py
+++ b/indexer/elastic.py
@@ -8,7 +8,7 @@ import argparse
 import os
 import sys
 from logging import getLogger
-from typing import Any, List, Mapping, Optional, Union
+from typing import Any, List, Optional
 from urllib.parse import urlparse
 
 from elastic_transport import NodeConfig, ObjectApiResponse
@@ -41,4 +41,5 @@ class ElasticMixin:
             logger.fatal("need --elasticsearch-hosts or ELASTICSEARCH_HOSTS")
             sys.exit(1)
 
+        # Connects immediately, performs failover and retries
         return Elasticsearch(hosts.split(","))
--- a/indexer/tests/test_importer.py
+++ b/indexer/tests/test_importer.py
@@ -8,7 +8,9 @@ from urllib.parse import urlparse
 
 import pytest
 from elastic_transport import NodeConfig
-from elasticsearch import Elasticsearch
+from elasticsearch import ConflictError, Elasticsearch
+
+from indexer.worker import QuarantineException
 
 # from indexer.elastic import create_elasticsearch_client
 from indexer.workers.importer import (
@@ -52,7 +54,7 @@ def elasticsearch_client() -> Any:
 test_data: Mapping[str, Optional[Union[str, bool]]] = {
     "original_url": "http://example.com",
     "normalized_url": "http://example.com",
-    "url": "http://example.com",
+    "url": "http://example40.com",
     "canonical_domain": "example.com",
     "publication_date": "2023-06-27",
     "language": "en",
@@ -64,6 +66,10 @@ test_data: Mapping[str, Optional[Union[str, bool]]] = {
     "indexed_date": datetime.now().isoformat(),
 }
 
+url = test_data.get("url")
+assert isinstance(url, str)
+test_id = hashlib.sha256(url.encode("utf-8")).hexdigest()
+
 
 class TestElasticsearchConnection:
     def test_create_index(self, elasticsearch_client: Any) -> None:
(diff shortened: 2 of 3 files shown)


indexer/elastic.py
```diff
@@ -42,4 +61,5 @@ def elasticsearch_client(self: ArgsProtocol) -> Elasticsearch:
             sys.exit(1)
 
         # Connects immediately, performs failover and retries
-        return Elasticsearch(hosts.split(","))
```
philbudne, line 45: This passes mypy for me, let's figure out why you're having problems and I'm not.

Honestly, I think duplicating code that exists inside the elasticsearch library and propagates a complex type signature is more of a long-term support liability than adding a comment to squelch mypy's complaints.

I had trouble satisfying the conditions of the big `Union`, and I think exposing that bit of horror should be avoided!!

pgulley (approved):

indexer/workers/importer.py
```diff
@@ -43,6 +44,7 @@
             "fields": {"keyword": {"type": "keyword"}},
         },
         "text_content": {"type": "text", "fields": {"keyword": {"type": "keyword"}}},
+        "indexed_date": {"type": "date"},
```
kilemensi, line 47: I could be misremembering but I believe this field shouldn't be searchable so we may need to turn [`indexed`](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-index.html) off... and since it's not part of `_source`, we may want to [`store`](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-store.html) it but not 100% sure on the last point.

kilemensi, line 47: I could be misremembering but I believe this field shouldn't be searchable so we may need to turn [`indexed`](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-index.html) off... and since it's not part of `_source`, we may want to [`store`](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-store.html) it but not 100% sure on the last point.

indexer/workers/importer.py
```diff
@@ -124,8 +126,9 @@ def index_routing(self, publication_date_str: Optional[str]) -> str:
                     year = -1
             except ValueError as e:
                 logger.warning("Error parsing date: '%s" % str(e))
-
-        index_name_prefix = self.index_name_prefix
+        index_name_prefix = self.index_name_prefix or os.environ.get(
+            "ELASTICSEARCH_INDEX_NAME_PREFIX"
+        )
```
kilemensi, line 131:
```suggestion
        index_name_prefix = self.index_name_prefix
```

[`os.environ`](https://github.com/mediacloud/story-indexer/blob/c19aea80e41abaca01db094d2014cce8046ffb7d/indexer/workers/importer.py#L90-L96) has already been used to set the default value here.

indexer/workers/importer.py
```diff
@@ -150,32 +153,38 @@ def process_story(
...
         """
         Import a single story to Elasticsearch
         """
         response = None
         if data:
+            url = str(data.get("url"))
+            url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
             publication_date = str(data.get("publication_date"))
+            # Add the indexed_date with today's date in ISO 8601 format
+            indexed_date: str = datetime.now().isoformat()
```
kilemensi, line 175: Doesn't `mypy` infer that `indexed_date` is a `str` from `isoformat` call automatically?

indexer/workers/importer.py
```diff
@@ -150,32 +153,38 @@ def process_story(
...
-                response = None
-                logger.error(f"Elasticsearch exception: {str(e)}")
+            except ConflictError as e:
+                self.incr(
+                    "imported-stories", labels=[("status", "409-duplicate-stories")]
+                )
+                raise QuarantineException("%s", str(e))
+            except RequestError as e:
+                self.incr("imported-stories", labels=[("status", "400-indexing-error")])
+                raise QuarantineException("%s", str(e))
```
kilemensi, line 187:
```suggestion
            except ConflictError as e:
                self.incr(
                    "imported-stories", labels=[("status", "409-duplicate-stories")]
                )
                raise QuarantineException from e
            except RequestError as e:
                self.incr("imported-stories", labels=[("status", "400-indexing-error")])
                raise QuarantineException from e
```
I think [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining) is better than casting if I understood the code.

indexer/workers/importer.py
rahulbot: The primary use case here is paging through new stories each night (described in #130), so we do want this field searchable because we'll need to filter for docs that have an `indexed_date` after some value. 

Perhaps you're thinking of `downloaded_date`, which we decided we don't need to index but do need to archive (see #135).

indexer/workers/importer.py
philbudne: @kilemensi -- I didn't write the QuarantineException catcher with chained exceptions in mind, so don't know how the code at https://github.com/mediacloud/story-indexer/blob/main/indexer/worker.py#L550 will handle your suggestion.

My comments on `QuarantineException("%s", str(e))` as submitted:
1. logger functions do % formatting, but QuarantineException is a class.
2. str(ExceptionSubclass("foo")) returns "foo", which is less useful than repr(ExceptionSubclass("foo")) which returns 'ExceptionSubclass("foo")', which gives the class name.

Not related to the exceptions, but in the above highlighted code is that I think the "imported-stories" counter should be changed to "stories", since the error events are stories that are NOT imported.

There usually isn't much room in the grafana graph windows (legend items are frequently not shown in full), so I try to keep names, label names and label values short, so I recommend using "dup" instead of "409-duplicate-stories".   Duplicates will happen regularly, so I don't think we want to collect them in quarantine.  We could start by doing it, but I think we would turn it off quickly.

For "400-indexing-error", I think "reqerr" is a reasonable short version of "RequestError", and if we don't understand all the ways this can happen, then collecting the stories in quarantine is a good idea.

indexer/elastic.py
philbudne, line 1: At this point I think elastic.py should be reset back to the original version: Mapping and Union are not being used, and the new version removes a comment (unless the comment is wrong)

indexer/tests/test_importer.py
```diff
@@ -59,8 +61,13 @@ def elasticsearch_client() -> Any:
     "article_title": "Example Article",
     "normalized_article_title": "example article",
     "text_content": "Lorem ipsum dolor sit amet",
+    "indexed_date": datetime.now().isoformat(),
 }
 
+url = test_data.get("url")
+assert isinstance(url, str)
+test_id = hashlib.sha256(url.encode("utf-8")).hexdigest()
+
```
kilemensi, line 70: Why are these here?

indexer/tests/test_importer.py
```diff
@@ -105,25 +132,35 @@ def test_import_story_success(
         elasticsearch_connector: ElasticsearchConnector,
     ) -> None:
         importer.connector = elasticsearch_connector
-        url = test_data.get("url")
-        assert isinstance(url, str)
-        id = hashlib.sha256(url.encode("utf-8")).hexdigest()
-        response = importer.import_story(id, test_data)
+        importer.index_name_prefix = os.environ.get("ELASTICSEARCH_INDEX_NAME_PREFIX")
```
kilemensi, line 135: Rather than setting `index_name_prefix` manually on every test, why not call `define_options`/`process_args` once in the test [`fixture`](https://github.com/mediacloud/story-indexer/blob/78285f40b99e2ffb4315f1501dbc29da1d4d5521/indexer/tests/test_importer.py#L120)?

indexer/workers/importer.py
```diff
@@ -150,32 +151,36 @@ def process_story(
...
             try:
                 response = self.connector.index(url_hash, target_index, data)
-            except Exception as e:
-                response = None
-                logger.error(f"Elasticsearch exception: {str(e)}")
+            except ConflictError:
+                self.incr("stories", labels=[("status", "dups")])
+
+            except RequestError as e:
+                self.incr("imported-stories", labels=[("status", "reqerr")])
```
kilemensi, line 182: Shouldn't `imported-stories` here also change to just `stories`?

philbudne (approved):
LGTM

thepsalmist:
@kilemensi any major blocker to merging this?

kilemensi:
> @kilemensi any major blocker to merging this?

I'll have to check @thepsalmist ... You didn't re-request review.

kilemensi (approved):

Issue #138, opened 2023-10-29
Alerting on logging errors


philbudne:
In https://github.com/mediacloud/story-indexer/issues/120 Rahul wrote:
> I'm a big fan of Sentry for capturing and surfacing warning/error logs from transient containers. Not the same as a whole log, but really useful and captures stack and state information as well. Also we use it across most of our other products. So a related issue is to integrate Sentry for remotely capturing ERROR log messages. Not sure I that should be a separate issue for discussion or is part of this.

#120 is about keeping mundane log messages across container lifetimes, so opening this one to track alerting on error/critical conditions.

philbudne:
NOTE: rss-fetcher sentry alerts are sent to the `#alerts-rss-fetcher` channel.
story-indexer should do similarly.

PR #147, opened 2023-11-08
add sentry for error logging

thepsalmist:
This PR introduces Sentry integration to capture error logs as in #138


philbudne (approved):

docker/deploy.sh
philbudne, line 1: The change to deploy.sh (pushed after my review) is incomplete:
1. it does not pass the deploy.sh environment variable setting to the JSON config file (using the "add" function)
2. the add function (on purpose) rejects empty string values, to detect unconfigured vars.  The "allow-null" (perhaps should be "allow-empty", since it isn't passing a JSON `null` value?) functionality exists, commented out in anticipation of such cases.
3. adding `add SENTRY_ENVIRONMENT` creates a `sentry_environment` value in config.json, which then needs to be substituted into worker container environment variable settings in docker-compose.yml.j2

I've just suggested (as opposed to my comments earlier today) in https://github.com/mediacloud/story-indexer/issues/148 that perhaps the settings for SENTRY_ENVIRONMENT and SENTRY_DSN values should both be done via the "private_vars" mechanism, since SENTRY_DSN is something we don't want to have in an open-source repository, and it would be clearer to have the two values (which are both deployment-type dependent) next to each other instead of being passed differently?!

indexer/sentry.py
```diff
@@ -0,0 +1,31 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def init() -> bool:
+    """
+    Centralized logging to Sentry.
+    """
+
+    sentry_dsn = os.environ.get("SENTRY_DSN")
```
philbudne, line 15: getting SENTRY_ENVIRONMENT is not needed, it's the default behavior of the sentry_sdk:
https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/client.py#L110

kilemensi:
I think this needs updating @thepsalmist

pgulley (approved):

thepsalmist:
@kilemensi this is ready for review, sourcing of the private env vars is already handled in `deploy.sh`. 
@rahulbot do we have the SENTRY_DSN generated to be set [here ](https://github.com/mediacloud/story-indexer-config/blob/main/prod.sh)

docker/deploy.sh
```diff
@@ -228,6 +229,7 @@ staging)
     MULTI_NODE_DEPLOYMENT=1
     NEWS_SEARCH_UI_TITLE="Staging $NEWS_SEARCH_UI_TITLE"
     VOLUME_DEVICE_PREFIX=/srv/data/docker/staging-indexer/
+    SENTRY_ENVIRONMENT="staging"
```
kilemensi, line 232: Is senty disabled in `dev`?

indexer/sentry.py
```diff
@@ -0,0 +1,29 @@
...
+            dsn=sentry_dsn,
+            # By default the SDK will try to use the SENTRY_RELEASE
+            # environment variable, or infer a git commit
+            # SHA as release, however you may want to set
+            # something more human-readable.
+            # release="indexer@1.0.0",
+        )
+        return True
+    else:
+        logger.warning("SENTRY_DSN not found. Not logging errors to Sentry")
```
kilemensi, line 28: I think it should be a warning in `prod` (or `staging`). Just info/debug in `dev`.

kilemensi (approved):

kilemensi:
> My vote: yes...
>
Totally with you on disabling sentry on dev @philbudne ... Otherwise, without explicitly setting it, `SENTRY_ENVIRONMENT` will default to [`production`](https://docs.sentry.io/platforms/python/configuration/options/#environment) if `SENTRY_DSN` is found and wreck havoc.

> Does it matter?  Seems like added complexity to lower the log level o
>
1. This statement was made with the assumption sentry **isn't** disabled in `dev` and hence it will **always** log a warning.
2. Correctness isn't a complexity... why say something is a warning when you 100% know it's not needed? (I'm assuming the fix here is an `if` statement).
3. I've approved the PR because I could be the only one who freaks out whenever I see error/warnings on my terminal.

philbudne:
> 2. Correctness isn't a complexity... why say something is a warning when you 100% know it's not needed? (I'm assuming the fix here is an `if` statement).
> 3. I've approved the PR because I could be the only one who freaks out whenever I see error/warnings on my terminal.

Well I don't want to shorten your life with unneeded stress!

The simple solution then, is to drop the message to "info" (or
"notice" if we add that).  Our only way to alert based on log messages
(at least at the current time) is via sentry, and if sentry isn't
properly configured, there is no way to alert, so the point seems
moot.

I've INTENTIONALLY avoided an environment setting that says this is
development, and instead aimed at having deploy.sh make the decisions
about what happens in what environment, rather than having many tiny
bit of policy decisions spread out in a thousand places.  In that
mental model, the "right" bit of configuration would be a variable
that says what level to log messages about missing sentry
configuration at, which is pretty circular, and getting/decoding that
would be a long/complex "if" statement.

A sidestep would be looking at the value of SENTRY_ENVIRONMENT, which,
is again, looking at SENTRY_ENVIRONMENT to decide how to complain
about the setting of SENTRY_DSN (again, seems circular) when they're
both set in the same place, on adjacent lines of code!

It feels a bit like betting your life on having two parachutes from
the same untrustworthy source.

Issue #134, opened 2023-10-25
Bandaids on long-running batches


pgulley:
Some batches take more than 24 hours to fetch. 
Long-term solution could be differently architected fetcher, but in the mean time:

1. [Download timout](https://docs.scrapy.org/en/latest/topics/settings.html#download-timeout) - Default is 2 minutes, we can probably tune that down
2. Redirect middleware- many of the pages in the longest running batch redirect to a domain which we will throw out anyway. Can we just catch that mid-stream instead?

3. We had discussed previously that batches over a certain size might just be ignored to begin with- we still might want to just filter at the batching step, presumably with the discretion of the directory team.

rahulbot:
For (1): I think 30 to 60 seconds is fine as a timeout. But that is a gut feeling, do we have any data on the breakdown of HTML response time across URLs to base a decision off of?

Latest chart from production shows distribution of runtime by day:
![story-indexer-all-realms](https://github.com/mediacloud/story-indexer/assets/673178/0c6a0a02-2e08-4f94-8c32-76bbc9bc8b82)

pgulley:
Looks like we'd need a custom redirect middle ware to stop requests in the middle of a redirect loop. Looks like a relatively simple lift, but it would be more code clutter in a component that iirc we may be thinking of rewriting wholesale, so wondering about the over-under on that.

rahulbot:
A fair point. Any way to quickly check the redirect history after content is fetched and filter there? Or is that equivalent to a check we're already doing?

pgulley:
Right now we're filtering on the final url, not the redirect history- so that might catch some small number more stories than we're currently catching- but it wouldn't address the issue of redirects taking too long or timing out.

philbudne:
I saw timeouts on fetching URLS from redirects in URLs from a now-disabled feed (archive.is redirecting to archive.ph, or visa versa)

Issue #127, opened 2023-10-23
Fetcher batches not completing


philbudne:
Even after splitting the daily rss-fetcher RSS file into 20 batches, one of the batches can fail to complete.  This seems to be (at least in part) due to archive.is (a source that is in two years' "top right wing sites" collections) feed containing many urls that redirect to archive.ph  (a domain we have in our "non-news" list) that time out when accessed.  If the fetcher process is killed before completing, NONE of the stories from that batch will be processed.

This can be ameliorated by https://github.com/mediacloud/story-indexer/issues/123
but it doesn't solve the problem (there may still be valid URLs that don't get fetched).

Maybe worth seeing if we can add scrapy middleware that detects redirects that go to domains we embargo, and stops the fetch (by throwing an exception, or returning a distinct HTTP status code?)
Again, this is papering over the problem.

rahulbot:
Band-aids in place. No problems in last 7 days. Still possible, but closing as not a priority.

We are separately tracking the idea of rewriting the fetcher to be queue-based (as opposed to Scrapy) on #128.
<img width="686" alt="story-indexer__all_realms__-_Grafana" src="https://github.com/mediacloud/story-indexer/assets/673178/0e763624-bf72-4821-9fc8-5e008adb23ec">

PR #150, opened 2023-11-09
Blacklist Redirect Middleware, and Shorter Timeout

pgulley:
Two changes- 
1. Add custom timeout, about half the duration of the default timeout.
2. Add a new Redirect Middleware which checks for non_news_domains in redirect urls before launching a redirect. 

Addressing issues in #134 (and #127)

### DIFF
--- a/indexer/workers/fetcher/BlacklistRedirectMiddleware.py
+++ b/indexer/workers/fetcher/BlacklistRedirectMiddleware.py
@@ -1,0 +1,50 @@
+from typing import Any
+from urllib.parse import urljoin, urlparse
+
+from mcmetadata.urls import NON_NEWS_DOMAINS
+from scrapy.downloadermiddlewares.redirect import (
+    BaseRedirectMiddleware,
+    _build_redirect_request,
+)
+from scrapy.exceptions import IgnoreRequest, NotConfigured
+from scrapy.http import HtmlResponse
+from scrapy.utils.httpobj import urlparse_cached
+from scrapy.utils.response import get_meta_refresh
+from w3lib.url import safe_url_string
+
+
+class BlacklistRedirectMiddleware(BaseRedirectMiddleware):  # type: ignore[no-any-unimported]
+    """
+    Handle redirection of requests based on response status
+    and meta-refresh html tag. Very Minor Edit of scrapy's builtin RedirectMiddleware
+    """
+
+    def process_response(self, request: Any, response: Any, spider: Any) -> Any:
+        if (
+            request.meta.get("dont_redirect", False)
+            or response.status in getattr(spider, "handle_httpstatus_list", [])
+            or response.status in request.meta.get("handle_httpstatus_list", [])
+            or request.meta.get("handle_httpstatus_all", False)
+        ):
+            return response
+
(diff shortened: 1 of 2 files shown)


indexer/workers/fetcher/BlacklistRedirectMiddleware.py
```diff
@@ -0,0 +1,50 @@
...
+            return response
+
+        location = safe_url_string(response.headers["Location"])
+        if response.headers["Location"].startswith(b"//"):
+            request_scheme = urlparse(request.url).scheme
+            location = request_scheme + "://" + location.lstrip("/")
+
+        redirected_url = urljoin(request.url, location)
+
+        if any(dom in redirected_url for dom in NON_NEWS_DOMAINS):
```
philbudne, line 42: Is this performing a substring search?  I had assumed NON_NEWS_DOMAINS (domain-skip-list.txt) would only ever contain a domain in the DNS sense, not the "canonical domain" sense of DNS_DOMAIN plus optional /prefix!! @rahulbot 

Doing substring search for `non_news_domain in full_url` (for each URL in a redirect chain) seems possible (even likely) that there could/will be spurious matches  if the path part of a URL ever contains a domain name!

I try to operate on the assumption that we live in an infinitely large and perverse universe/Internet, and that any bad thing WILL eventually happen, and that it will have some (eventually) noticeable bad effect!

Also: A string search inside a loop, executed for each redirect step is O(n^3), tho none of the loops is THAT large.
A (possibly premature optimization, which we know to be the root of all evil) would be to parse the domain out of the redirected_url and look it up in a precomputed set of (DNS) domains found in NON_NEWS_DOMAINS.

rahulbot: I think `extracted_domain in NON_NEWS_DOMAINS` in a DNS sense sounds right (or `canonicaldomain.endswith` for each one). Doing an `in redirected_url` check also sets my spidey sense tingling about spurious matches, like it does for Phil.

pgulley: I believe that the values in [NON_NEWS_DOMAINS](https://github.com/mediacloud/metadata-lib/blob/main/mcmetadata/data/domain-skip-list.txt) are already "website.com" (ie including the tld) rather than just domain names, which I think makes spurious matches a _very_ perverse edge-case. But extracting the domain would safeguard against this and make the check less expensive, so might as well?

philbudne:
News story URL paths very often contain something like
a-concatenation-of-headline-words, so it's not terribly far fetched
that story headlines might contain "full" names of popular, but
excluded domains like linkedin.com, tiktok.com, youtube.com or
wikipedia.org which might be news-worthy.

indexer/workers/fetcher/BlacklistRedirectMiddleware.py
pgulley: I've made a change to use the tldextract package so we're checking against the actual domain instead of searching through the string.

philbudne (approved):

kilemensi (approved):

thepsalmist (approved):
LGTM

PR #151, opened 2023-11-10
Reconfigure fetch worker to use new on-the-fly queueing via a pika thread

pgulley:
As written

### DIFF
--- a/indexer/workers/fetcher/fetch_worker.py
+++ b/indexer/workers/fetcher/fetch_worker.py
@@ -29,7 +29,7 @@ MAX_FETCHER_MSG_SIZE: int = (
 
 
 class FetchWorker(StoryProducer):
-    AUTO_CONNECT: bool = False
+    START_PIKA_THREAD: bool = True
 
     fetch_date: str
     sample_size: Optional[int]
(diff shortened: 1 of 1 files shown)


philbudne:
Looking at this, I suspect you made exactly the changes I suggested
(replacing `AUTO_CONNECT = False` with `START_PIKA_THREAD = True`)
but the batching/StorySender changes introduced new complications!!

Of the two warnings in StorySender.story_sender:

1. Call once PER THREAD
	I'm not sure what my concern was here..
	(may have been excess paranoia, since
	all the channel access is done in the context
	of the Pika thread!)

2. MUST be called before Pika thread running

	This is the concern: Once the Pika thread is running, the
	connection belongs to it, and calling
	self.connection.channel() is asking for trouble.

So I NOW think:
Remove setting `START_PIKA_THREAD = True` and...

Since scrapy is single threaded, I think doing the connect lazily
the first time a story needs to be sent is safe:

    if self.sender is None:
        self.qconnect()
	self.sender = self.story_sender()
    self.sender.send_story(story)

The alternative would be, at the top of main_loop
1. call self.qconnect()
2. self.sender = self.story_sender()
3. call self._start_pika_thread()

I've created issue
https://github.com/mediacloud/story-indexer/issues/152 and assigned
myself to do cleanup around this

pgulley:
Will move on updating and testing this this week, now that #152 is addressed.

pgulley:
Replaced by #187

PR #162, opened 2023-11-17
Proposed changes to deployment variables for elastic-snapshots service.

philbudne:
The point of this change set is to give more flexibility on when the elastic-snapshots container is launched.

0. Enhance "int" in deploy.sh to handle negative numbers
1. Add ELASTICSEARCH_SNAPSHOT_CONTAINERS (int)
     if < 0: shapshots disabled (default/dev)
     if = 0: do not launch when stack deployed (production)
     if > 0: launch at stack deployment
2. Add ELASTICSEARCH_SNAPSHOT_REPO (str)
3. Add ELASTICSEARCH_SNAPSHOT_ARGS (allow-empty) passed in ARGS (to allow extra args in staging, if needed, to define shapshot, load keys)
5. Change cronjob to run at 01:00GMT (allow time to redeploy stack after 00:00GMT)
6. Add swarm.cronjob.replicas: 1 (for production)

### DIFF
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -106,7 +106,7 @@ fi
 ELASTICSEARCH_CLUSTER=mc_elasticsearch
 ELASTICSEARCH_IMAGE="docker.elastic.co/elasticsearch/elasticsearch:8.8.0"
 ELASTICSEARCH_PORT_BASE=9200	# native port
-ELASTICSEARCH_SNAPSHOT_CRONJOB_ENABLE=false
+ELASTICSEARCH_SNAPSHOT_CONTAINERS=-1 # disable container by default
 
 FETCHER_CRONJOB_ENABLE=true
 FETCHER_NUM_BATCHES=20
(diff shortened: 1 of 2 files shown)


kilemensi (approved):

docker/deploy.sh
```diff
@@ -200,10 +200,10 @@ prod)
     ELASTICSEARCH_IMPORTER_SHARDS=30
 
     # Disabled until tested in staging.
-    # Questions:
-    # 1. needs additional config?
-    # 2. will restart each time docker-compose deploy run??
+    # Disable starting at deployment, run periodically:
+    #ELASTICSEARCH_SNAPSHOT_CONTAINERS=0
     #ELASTICSEARCH_SNAPSHOT_CRONJOB_ENABLE=true
+    ELASTICSEARCH_SNAPSHOT_REPO=mediacloud-elasticsearch-snapshots
```
thepsalmist, line 206: LGTM

Just checked registered repo-name should be `mc_ec_s3_repository`

`{
    "mc_ec_s3_repository": {
        "type": "s3",
        "settings": {
            "bucket": "mediacloud-elasticsearch-snapshots"
        }
    }
}`

thepsalmist: While on this, was thinking if we can actually use [Elasticsearch's SLM ](https://www.elastic.co/guide/en/elasticsearch/reference/current/snapshots-take-snapshot.html#automate-snapshots-slm)to manage the scheduling instead of relying on the external Cron.

This should be non-blocking, if it makes sense we can add the script to this or open an issue around it

pgulley (approved):

docker/deploy.sh
```diff
@@ -217,6 +217,13 @@ staging)
     ELASTICSEARCH_IMPORTER_REPLICAS=1
     ELASTICSEARCH_IMPORTER_SHARDS=5
 
+    # Disabled until snapshot configuration & keys set.
+    # (pass keys in environment, set ELASTICSEARCH_SNAPSHOT_ARGS w/ option?)
+    # Start at deployment, do not run periodically:
+    #ELASTICSEARCH_SNAPSHOT_CONTAINERS=1
+    #ELASTICSEARCH_SNAPSHOT_CRONJOB_ENABLE=false
```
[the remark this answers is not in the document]
philbudne: My idea here was maximum flexibility.  Is there a plan that doesn't require a cronjob?

Issue #180, opened 2023-12-04
importer duplicate handling


philbudne:
As seen in slack: https://app.slack.com/client/T0SRSUKKM/C03TE8NA675

Paige deployed new code to prod (at about 18:45 GMT?), which should have started fetching the same stories as we fetched yesterday, so it's interesting to see what's happening:

![image](https://github.com/mediacloud/story-indexer/assets/7842179/86aa8163-745c-4894-87a5-28bfbffc4f63)

It looks like:
1. importer is reporting most stories as "failed" and not "dups"
2. that the duplicate stories are being passed to archiver, which duly writes archive files
3. the index didn't grow (good)

My opinions:
1. Duplicate stories should be discarded (not passed to archiver)
2. Duplicate stories should be reported as duplicate (OK to have more than one counter for this: rss-fetcher does!)
3.  status="failed" should be reserved for unknown failure conditions, and
4. "failed" stories should be quarantined for later examination.

rahulbot:
I agree with those suggested behaviors.

Issue #155, opened 2023-11-13
ES rejecting stories with text_content field over 32KB


philbudne:
Got 1.9K over two days (650K stories imported OK), so 0.3%
As opposed to parser quarantine which got 8.1K (1.25%)

rahulbot:
Not sure how to prioritize this info... is the next task to asses the frequent causes of quarantine in each? 
Ie. (a) bugs, (b) cases to retry, or (c) cases that are failures we want to toss out of the pipeline, etc.

philbudne:
Examining almost 10K quarantined messages in pbudne-indexer stack;
when dumped using `indexer.scripts.qutil dump_archives` 100% of them
have saved headers like:
```
  "rabbitmq_headers": {
    "x-mc-who": "importer",
    "x-mc-when": "1703472865.3358805",
    "x-mc-what": "QuarantineException('document_parsing_exception')",
    "x-mc-where": "/app/indexer/workers/importer.py:182",
    "x-mc-name": "import_story"
  }
```
Which come from:
```
           except RequestError as e:
                self.incr("stories", labels=[("status", "reqerr")])
                raise QuarantineException(getattr(e, "message", repr(e)))
```
(one of the two files is available as ~pbudne/importer-quar/importer-quar-20240111053147-2-ifill.warc.gz

It's preferable to do `raise QuarantineError(repr(e))`; using just e.message discards information (what type of exception was caught, requiring examining the right version of the source code).

philbudne:
With latest changes to importer.py, "x-mc-what" is now
`QuarantineException('BadRequestError(\\'document_parsing_exception\\', meta=ApiResponseMeta(status=400",`

philbudne:
As seen on slack:

I just did a binary search on the text of a Story that had gotten a document parse error on import (we currently have 71.6k stories in the importer quarantine queue).  It turns out that if the text_content field is 32766 bytes long when utf-8 encoded it succeeds, and if it's 32767 bytes, it fails.  32767 is 2^15-1, which has the smell of length being limited a 15 bit (or signed 16 bit) field.  Googling found https://discuss.elastic.co/t/how-big-a-field-can-be/158810

philbudne:
My (possibly feeble) understanding is that the fixes are all unattractive:
1. simply discard the stories
2. truncate the text before attempting import
3. store the text in multiple fields (which would explode query expression complexity)
4. tell ES to ignore characters above the limit (store them, but not searchable?)
[...]

PR #189, opened 2023-12-08
importer bug fixes

thepsalmist:
Bug Fixes as highlighted in #180

### DIFF
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -13,7 +13,6 @@ from elastic_transport import ObjectApiResponse
 from elasticsearch import Elasticsearch
 from elasticsearch.exceptions import ConflictError, RequestError
 
-from indexer.app import AppException
 from indexer.elastic import ElasticMixin
 from indexer.story import BaseStory
 from indexer.worker import QuarantineException, StorySender, StoryWorker, run
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -9,10 +9,11 @@ import sys
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Mapping, Optional, Union, cast
 
-from elastic_transport import NodeConfig, ObjectApiResponse
+from elastic_transport import ObjectApiResponse
 from elasticsearch import Elasticsearch
 from elasticsearch.exceptions import ConflictError, RequestError
 
+from indexer.app import AppException
 from indexer.elastic import ElasticMixin
 from indexer.story import BaseStory
 from indexer.worker import QuarantineException, StorySender, StoryWorker, run
@@ -151,7 +152,6 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
                 k: v for k, v in content_metadata.items() if k not in keys_to_skip
             }
             self.import_story(data)
-
             # pass story along to archiver
             # (have an option to disable, for previously archived data?)
             sender.send_story(story)
@@ -174,8 +174,12 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
             target_index = self.index_routing(publication_date)
             try:
                 response = self.connector.index(url_hash, target_index, data)
-            except ConflictError:
+            except ConflictError as e:
                 self.incr("stories", labels=[("status", "dups")])
+                # raise exception so that dups are not archived
+                raise AppException(
+                    "Conflict error occurred during indexing (duplicates detected)"
+                ) from e
 
             except RequestError as e:
                 self.incr("stories", labels=[("status", "reqerr")])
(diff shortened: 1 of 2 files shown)


pgulley (approved):

indexer/workers/importer.py
```diff
@@ -174,8 +174,12 @@ def import_story(
             target_index = self.index_routing(publication_date)
             try:
                 response = self.connector.index(url_hash, target_index, data)
-            except ConflictError:
+            except ConflictError as e:
                 self.incr("stories", labels=[("status", "dups")])
+                # raise exception so that dups are not archived
```
philbudne, line 179: raising Exceptions (other than QuarantineException) from process_story will cause the story to be retried (a fixed number of times) before the story is thrown to the quarantine queue.  Also, the exception handler in indexer.worker doesn't know about the `raise ... from` construct.  It may (or may not) cause useful information to be put into the queue message headers for later diagnosis.

If the goal is to not pass the story onto the archiver, the correct thing to do is have `process_story` return without calling `sender.send_story`

But, I think what I saw in the graphs in Issue #180 is that duplicates are being counted as "failed", and not showing up in this code path, but rather the one at new line 195 (old line 192) below.

indexer/workers/importer.py
```diff
@@ -187,9 +191,10 @@ def import_story(
                 )
                 import_status_label = "success"
             else:
-                # Log no imported stories
                 logger.info("Story was not imported.")
                 import_status_label = "failed"
```
philbudne, line 195: My reading of the graphs in #180 is that duplicate insert attempts are being counted as "failed", so code is needed here to distinguish if that is the case (or whether some other error occurred).

Note: Having this code in a sub-method (import_story) that's called only once makes it harder to "just return from process_story" without throwing an exception or, (gasp) returning status information, so inlining the method might make things simpler.

Finally, if the error is something other than a duplicate, it would be useful to have information in the QuarantineException that helps diagnose what might have caused the error (ie; a repr of the exception or of the response).

philbudne (changes requested):

indexer/workers/importer.py
thepsalmist: On the graphs in #180, the PR already provides for three status for failed imports `dups` for duplicates, `reqerr` for 403 errors else `failed`. This should have been displaying on the grafana dashboard, but we hadn't[ updated this](https://github.com/mediacloud/story-indexer/blob/476ea3487c64b9b731a28da45280b1c7b9352a07/conf/grafana/story-indexer-all-realms.json#L419) to a uniform `importer.stories`

kilemensi (approved):

indexer/workers/importer.py
philbudne: Note: the current state of affairs is that the The dashboard files in git been downloaded from grafana (where they are currently authored) and checked in to git preserve them:
1. changes may have been made in the dashboards that haven't yet been exported and checked in to git
2. changes made in git will NOT appear in grafana

philbudne: Manually switching the view to "stories" gives this (does not show successes?):
image

philbudne: A good way to convince me the changes do the right thing is to run a developer instance and capture the (new) grafana output!

philbudne (commented):
Looking directly at the thepsalmist:fix/importer-dup-handling branch (for net effect of PR):

1. It's hard (for me) to tell that each Story will increment exactly one "status" value: what if response is `None` or response.result != "created"?
2. `raise QuarantineException(getattr(e, "message", repr(e)))` discards information on what exception was the cause; see https://github.com/mediacloud/story-indexer/issues/155
3. Catching `Exception` and quarantining means EVERY unhandled error (possibly including ES server unreachable?) is immediately quarantined, instead of retried.  It would be preferable to re-raise (ie; bare `raise`) any unidentified (ie; not KNOWN to be a transient problem) problem, so the Story is retried a number of times before ending up quarantined.

philbudne (commented):
This seems much clearer, but:
1. please don't make changes to the .json files (see the README.md now in main) `The files here have been exported for safe keeping and should not be edited.`
2. in ElasticsearchImporter.import_story: if ElasticsearchConnector.index returns a response object where `response.get("result") != "created"`,a non-None response is returned (story is passed to archiver), and the story is archived, but NOT counted, presumably hasn't been saved in ES, and is neither retried (until finally quarantined), nor immediately quarantined.

philbudne:
Looking at the ES API docs: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html

The only listed values for `result` are `created` and `updated`.

And as a RESTful API, it seems to signal errors via HTTP response codes, which are turned into exceptions by the ES library.

So perhaps the right thing to do is use the `result` value as the status tag to use to increment the counter?  But what to do if no result value is available?  Maybe use `noresult`?

Putting the response check and return inside the `try:` might make it more clear that it only applies
when no exception is raised, and put "return None" (or False, if you return a boolean) at the end???

philbudne (approved):

Issue #184, opened 2023-12-06
parser html decode: discard or replace?


philbudne:
At our November 2023 in-person meeting @rahulbot and I agreed that we should try `bytes.decode(encoding, errors='ignore')`

since then I prowled the legacy system one more time and found:

https://github.com/mediacloud/backend/blob/master/apps/common/src/python/mediawords/util/web/user_agent/response/response.py#L192

```
            decoded_content = codecs.decode(self.__response_data, encoding=encoding, errors='replace')
```

'replace' means "Replace with a suitable replacement character" on decode error.

It looks like "suitable replacement" is unicode FFFD "REPLACEMENT CHARACTER" which displays as � (which is what, I think, most browsers do).

rahulbot:
Sounds good. No point in throwing away content when we can get some version of it that _might_ be useful but using this approach you outline (and looks like we used to do in legacy).

philbudne:
Just tried this (in the Story object html property method):
```
-            return self.html.decode(self.encoding)
+            return self.html.decode(self.encoding, errors='replace')
```

And was able to reduce  the quarantine queue from the pbudne-staging environment from 1465 messages, down to 20 (with "no html")

@pgulley does this seem like a reasonable place to put the "fix"

PR #191, opened 2023-12-10
use errors="replace" in RawHTML html property method

philbudne:
'replace' means "Replace with a suitable replacement character" on decode error (was causing quarantine).  It looks like "suitable replacement" is unicode code point FFFD "REPLACEMENT CHARACTER" which displays as � (which is what, I think, most browsers do).

fixes https://github.com/mediacloud/story-indexer/issues/184


kilemensi (approved):
🚀

pgulley (approved):

Issue #186, opened 2023-12-07
importer: many WARNINGS on published_date "None"


philbudne:
Every(?) missing published_date is logged as a warning by the importer:
`Error parsing date: 'time data 'None' does not match format '%Y-%m-%d'`

Perhaps `if publication_date_str:`  in the`index_routing` function should be `if publication_date_str and publication_date_str != 'None':`??

philbudne:
Since we have to re-index all data, perhaps we could store unknown publication data as Python `None` / JSON `null` instead of the string `"None"`???

pgulley:
I'll move on this asap

philbudne:
This issue will be mooted by https://github.com/mediacloud/story-indexer/pull/211 since the code that was logging the error will have gone away.

Issue #172, opened 2023-12-01
use story publication date from RSS as fallback if none is found


rahulbot:
If no publication date is found when parsing the story, but the story URL is from RSS and it had a publication date on that RSS entry, then use that RSS publication date.

This can be accomplished via the new `defaults` param supported by the `metadata.extract` method (once `mc-metadata >= 0.10.0` is released)

rahulbot:
(v0.10.0 of mc-metadata is now released and this new feature is available)

philbudne:
The new code is:
```
            # if publication date is none, fallback to rss_fetcher pub_date
            if data["publication_date"] is None:
                data["publication_date"] = story.rss_entry()["pub_date"]
```

However looking at metadata in archive WARC files content metadata "publication_date" format is `YYYY-MM-DD`, while rss entry pub_date seems to be `Dow, DD Mon YYYY HH:MM:SS TZ` (ie; RFC 2822) and is shown in this format in https://www.rssboard.org/rss-specification#ltpubdategtSubelementOfLtitemgt

https://www.w3schools.com/xml/rss_tag_pubdate_item.asp shows `Dow, DD Mon YYYY`

PR #208, opened 2024-01-18
pub date falls back to rss_fetcher date when missing- skipping mypy p…

pgulley:
…recommit due to local dependency issue

Dev instance of the pipeline still runs without errors

### DIFF
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -184,7 +184,10 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
             url = str(data.get("url"))
             url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
             # We want actual None, not 'None', if publication_date is missing
-            if "publication_date" in data:
+            if "publication_date" in data and data["publication_date"] not in [
+                None,
+                "None",
+            ]:
                 publication_date = str(data["publication_date"])
             else:
                 publication_date = None
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -165,13 +165,12 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
 
             # if publication date is none, fallback to rss_fetcher pub_date
             if data["publication_date"] is None:
-                data["publication_date"] = story.rss_entry().as_dict()["pub_date"]
+                data["publication_date"] = story.rss_entry()["pub_date"]
 
             response = self.import_story(data)
             if response and self.output_msgs:
                 # pass story along to archiver (unless disabled)
                 sender.send_story(story)
-
 
     def import_story(
         self,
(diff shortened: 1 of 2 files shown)


pgulley:
For #186 and #172

indexer/workers/importer.py
```diff
@@ -158,14 +158,21 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
...
 
             keys_to_skip = ["is_homepage", "is_shortened"]
+
             data: Mapping[str, Optional[Union[str, bool]]] = {
                 k: v for k, v in content_metadata.items() if k not in keys_to_skip
             }
+
+            # if publication date is none, fallback to rss_fetcher pub_date
+            if data["publication_date"] is None:
+                data["publication_date"] = story.rss_entry().as_dict()["pub_date"]
```
kilemensi, line 168: Doesn't `story.rss_entry().pub_date` return the `pub_date` without converting everything to `dict` first?

indexer/workers/importer.py
```diff
@@ -177,7 +184,11 @@ def import_story(
         if data:
             url = str(data.get("url"))
             url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
-            publication_date = str(data.get("publication_date"))
+            # We want actual None, not 'None', if pub_date is missing
+            if (pub_date := data.get("publication_date")) is None:
+                publication_date = None
+            else:
+                publication_date = str(pub_date)
```
kilemensi, line 191:
```suggestion
            # We want actual None, not 'None', if publication_date is missing
            if ("publication_date" in data) is None:
                publication_date = str(data["publication_date"])
            else:
                publication_date = None
```

kilemensi (commented):
👍🏽 

--

Shouldn't this PR address [this comment](https://github.com/mediacloud/story-indexer/blob/6f90ae7fb2467d1ea4d820b1c78db3a706a27124/indexer/workers/parser.py#L123) as well?

indexer/workers/importer.py
```diff
@@ -185,7 +184,7 @@ def import_story(
             url = str(data.get("url"))
             url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
             # We want actual None, not 'None', if publication_date is missing
-            if ("publication_date" in data):
+            if "publication_date" in data:
```
philbudne, line 187: Can "publication_date" item be present, but `None` or `"None"` (especially when replaying WARC files)??

pgulley: Good point, i've addressed this in the recent commit

philbudne (approved):
Looks good.  `data.get("publication_date") not in [None, "None"]` would be shorter

kilemensi (approved):
🚀

PR #212, opened 2024-01-23
update RSS date

thepsalmist:
Quickfix pub_dates falling to RSS pub_dates if None

### DIFF
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -9,7 +9,6 @@ import sys
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Mapping, Optional, Union, cast
 
-from dateutil import parser, tz
 from elastic_transport import ObjectApiResponse
 from elasticsearch import Elasticsearch
 from elasticsearch.exceptions import ConflictError, RequestError
@@ -160,7 +159,7 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
 
             keys_to_skip = ["is_homepage", "is_shortened"]
 
-            data: Mapping[str, Optional[Union[str, bool]]] = {
+            data: Dict[str, Optional[Union[str, bool]]] = {
                 k: v for k, v in content_metadata.items() if k not in keys_to_skip
             }
 
@@ -168,8 +167,9 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
             if data["publication_date"] is None:
                 pub_date = story.rss_entry().pub_date
                 if pub_date:
-                    es_date = parser.parse(pub_date)
-                    data = {**data, "publication_date": es_date}
+                    date_obj = datetime.strptime(pub_date, "%a, %d %b %Y %H:%M:%S %z")
+                    es_date = date_obj.isoformat()
+                    data["publication_date"] = es_date
 
             response = self.import_story(data)
             if response and self.output_msgs:
@@ -178,7 +178,7 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
 
     def import_story(
         self,
-        data: Mapping[str, Optional[Union[str, bool]]],
+        data: Dict[str, Optional[Union[str, bool]]],
     ) -> Optional[ObjectApiResponse[Any]]:
         """
         Import a single story to Elasticsearch
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -168,8 +168,7 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
             if data["publication_date"] is None:
                 pub_date = story.rss_entry().pub_date
                 if pub_date:
-                    tzinfos = {"UTC": tz.UTC} 
-                    es_date = parser.parse(pub_date, tzinfos=tzinfos).isoformat()
+                    es_date = parser.parse(pub_date)
                     data = {**data, "publication_date": es_date}
 
             response = self.import_story(data)
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -7,7 +7,7 @@ import logging
 import os
 import sys
 from datetime import datetime, timedelta
-from typing import Any, Dict, List, Mapping, Optional, Union, cast
+from typing import Any, Mapping, Optional, Union, cast
 
 from elastic_transport import ObjectApiResponse
 from elasticsearch import Elasticsearch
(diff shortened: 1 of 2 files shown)


indexer/workers/importer.py
```diff
@@ -165,7 +166,11 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
 
             # if publication date is none, fallback to rss_fetcher pub_date
             if data["publication_date"] is None:
-                data["publication_date"] = story.rss_entry()["pub_date"]
+                pub_date = story.rss_entry().pub_date
+                if pub_date is not None and pub_date != "":
```
kilemensi, line 170:
```suggestion
                if pub_date:
```
Should capture both `None` and `""`

indexer/workers/importer.py
```diff
@@ -165,7 +166,11 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
...
+                    tzinfos = {"UTC": tz.UTC} 
+                    es_date = parser.parse(pub_date, tzinfos=tzinfos).isoformat()
```
kilemensi, line 172:
```suggestion
                    es_date = parser.parse(pub_date)
```
`parse` should understand `UTC`, `GMT`, `Z` by [default](https://dateutil.readthedocs.io/en/stable/parser.html#classes).
According to Phil's comment [here](https://github.com/mediacloud/story-indexer/issues/172#issuecomment-1902514227), it's possible for `pub_date` to have a missing timezone. We may need to check if `tzinfo` exists before calling `isoformat`

indexer/workers/importer.py
```diff
@@ -165,7 +166,11 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
...
+                    data = {**data, "publication_date": es_date}
```
kilemensi, line 173: It's not a big deal but isn't `data["publication_date"] = ...` better than making a copy of the whole `dict`? We're only interested in updating one field after all.

kilemensi (commented):
👍🏽

indexer/workers/importer.py
thepsalmist: Yes ideally, the issue with that arose because the dictionary data is of type `Mapping[str, str | bool | None]`, which is an immutable mapping type

philbudne: Since we've postponed the re-index, I'd been leaving this for Paige to deal with, and wondering if we should just roll back the problem commit??

re Mapping: AH! so that's why I keep seeing copying!  Any reason to use `Mapping` instead of `Dict` (isn't Dict is a kind of Mapping? or use `MutableMapping`??)

re dateutil:

1. The date comes from rss-fetcher, and will always be RFC-(2)822 format (formatted by email.utils.formatdate), and strptime with `'%a, %d %b %Y %H:%M:%S %z'` (or similar) should always suffice, and dateutil is unnecessary.
2. as a general point: "parse" isn't particularly self-descriptive, so in cases like this, just importing the module and using `dateutil.parser.parse` is less confusing.  It looks like `tz` isn't used (I'd _really_ like to reenable flake8 complaints about

thepsalmist: The initial use case foresaw only reading `content_metadata` without any need to Mutate, hence `Mapping`. Converting this to `Dict`

indexer/workers/importer.py
```diff
@@ -159,13 +159,17 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
 
             keys_to_skip = ["is_homepage", "is_shortened"]
 
-            data: Mapping[str, Optional[Union[str, bool]]] = {
+            data: Dict[str, Optional[Union[str, bool]]] = {
```
kilemensi, line 162: Since python [`3.9`](https://peps.python.org/pep-0585/), I don't think there is any benefit of using `typing.Dict` instead of just `dict`.

kilemensi (approved):

indexer/workers/importer.py
```diff
@@ -159,13 +159,24 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
...
+            if pub_date is None or pub_date == "None":
+                rss_pub_date = story.rss_entry().pub_date
+                if rss_pub_date:
+                    es_pub_date = datetime.strptime(
+                        rss_pub_date, "%a, %d %b %Y %H:%M:%S %z"
+                    ).strftime("%Y-%m-%d")
+                else:
+                    es_pub_date = None
+                if pub_date != es_pub_date:
+                    data["publication_date"] = es_pub_date
```
kilemensi, line 177: ### 
```suggestion
            pub_date = data["publication_date"]
            if pub_date in [None, "None"]:
                rss_pub_date = story.rss_entry().pub_date
                if rss_pub_date:
                    pub_date = datetime.strptime(rss_pub_date, "%a, %d %b %Y %H:%M:%S %z").strftime("%Y-%m-%d")
                else:
                    pub_date = None
                data["publication_date"] = pub_date
```

Assuming
1. `data["publication_date"]` can't be empty (`""`) at this point; just `None` or `"None"`
2. if `rss_pub_date` exists, `strptime` will never throw `ValueError`.

kilemensi (approved):

indexer/workers/importer.py
```diff
@@ -183,18 +194,11 @@ def import_story(
...
-            else:
-                publication_date = None
-            # Add the indexed_date with today's date in ISO 8601 format
-            indexed_date = datetime.now().isoformat()
-            data = {**data, "indexed_date": indexed_date}
-            target_index = self.index_routing(publication_date)
+            # XX This wont't be needed for Elasticsearch ILM -To remove
+            publication_date = data.get("publication_date")
+            # To move to Story index metadata
+            data["indexed_date"] = datetime.utcnow().strftime("%Y-%m-%d")
```
kilemensi, line 200: Based on this [clarification](https://github.com/mediacloud/story-indexer/issues/213#issuecomment-1908906459) from Rahul, `indexed_date` must be remain date+time as is now in the [`main`](https://github.com/mediacloud/story-indexer/blob/208978787d5978ffa210b9e7718a155108f5d3e5/indexer/workers/importer.py#L195) branch.

indexer/workers/importer.py
thepsalmist: > if rss_pub_date exists, strptime will never throw ValueError

from rss-fetcher, and will always be RFC-(2)822 format (formatted by email.utils.formatdate)

thepsalmist:
Merged to update dependencies on #211

Issue #213, opened 2024-01-24
ES indexed_date field when processing WARC files, and historical data


philbudne:
In today's (2023-01-24) team meeting, I got the understanding from @rahulbot that ES `indexed_date` is meant to indicate "when the data was processed (ie; parsed)"

In the meeting, I thought `ContentMetadata` contained a timestamp, but it turns out I was thinking of `HTTPMetadata.fetch_timestamp`

For the upcomming re-import/re-index from WARC files some choices are:
1. add `ContentMetadata.parse_timestamp` field (a small change), have parser.py always set it (to `time.time()`) and have the archive processing process (arch-queuer.py) default it to the `WARC-Date` if the field is missing when reading archives, and have the importer use ContentMetadata.parse_timestamp to generate `indexed_date` from it, or default to `datetime.utcnow()`
2. not add the field, fill in `HTTPMetadata.fetch_timestamp` and pass the importer a command line option to use fetch_timestamp instead of "now" for generation of `indexed_date`

For processing historical data, some choices are:
1. same as 1 above: since the data will pass thru the parser, the indexed date will indicate the date it was parsed.
2. same as 2 above, except have `arch-fetcher.py` fill in `HTTPMetadata.fetch_timestamp` with `time.time()` instead of either the timestamp of the S3 object containing the archived HTML, or the date/time in the CSV file.   BUT, the HTTP data date could be saved in the `RSSEntry` Story sub-object, but this feels like a terrible crock!

philbudne:
In both cases above one more option is:
3. do nothing, have `indexed_date` reflect the date the data was entered into ES.

My understanding is that this is not a terrible choice (because it wouldn't be off the parse date by more than 70 days), BUT would be a less ideal choice at some future time, because the indexed date would be off by some large amount)

rahulbot:
Thank you for the clear summary. It avoids some other downstream headaches I'm thinking of if we have an `indexed_date` on these re-indexed stories that is close to accurate. For that reason for the upcoming re-indexing **I prefer option 1**, which looks to me to be more involved but still clear. The responsibilities for handling are well-divided across jobs in your description. 

[...]

PR #219, opened 2024-01-26
Add "parsed_date" field to content metadata

philbudne:
New "parsed_date" field allows passing and preserving the date the story was processed, for use as ES "indexed_date".

This is required to queue archived stories and have them put back into the index preserving original date
(API users use "indexed_date" to poll for new stories).

Code might be _slightly_ less complicated if field name was "indexed_date", but that would obscure the actual meaning of the field!

### DIFF
--- a/indexer/story_archive_writer.py
+++ b/indexer/story_archive_writer.py
@@ -364,6 +364,13 @@ class StoryArchiveReader:
                 with story.content_metadata() as cmd:
                     for key, value in j["content_metadata"].items():
                         setattr(cmd, key, value)
+                    if not cmd.parsed_date:
+                        # For archives written before "parsed_date" added;
+                        # Use WARC metadata record WARC-Date header.
+                        date = record.rec_headers["WARC-Date"]
+                        if date.endswith("Z"):  # should always be the case
+                            date = date[:-1]  # remove Z
+                        cmd.parsed_date = date
                 with story.raw_html() as rh:
                     rh.html = html
                     rh.encoding = j["http_metadata"]["encoding"]
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -179,8 +179,11 @@ class ElasticsearchImporter(ElasticMixin, StoryWorker):
                 with story.content_metadata() as cmd:
                     cmd.publication_date = pub_date
 
-            # pass parsed_date (from parser or an archive file) as indexed_date
-            # fall back to UTC now.
+            # Use parsed_date (from parser or an archive file) as indexed_date,
+            # falling back to UTC now (for everything parsed/queued before
+            # update applied to parser). API users use indexed_date to poll for
+            # newly added stories, so a timestamp.  By default, ES stores times
+            # with millisecond granularity.
             data["indexed_date"] = (
                 content_metadata.get("parsed_date") or datetime.utcnow().isoformat()
             )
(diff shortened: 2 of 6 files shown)


philbudne:
fixes https://github.com/mediacloud/story-indexer/issues/213

kilemensi (approved):
🚀

indexer/story_archive_writer.py
```diff
@@ -364,6 +364,13 @@ def read_stories(self) -> Iterator[BaseStory]:
                 with story.content_metadata() as cmd:
                     for key, value in j["content_metadata"].items():
                         setattr(cmd, key, value)
+                    if not cmd.parsed_date:
+                        # For archives written before "parsed_date" added;
+                        # Use WARC metadata record WARC-Date header.
+                        date = record.rec_headers["WARC-Date"]
```
rahulbot, line 370: Is there any chance `date` could be None here that needs to be handled?

indexer/workers/importer.py
```diff
@@ -179,6 +179,14 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
                 with story.content_metadata() as cmd:
                     cmd.publication_date = pub_date
 
+            # Use parsed_date (from parser or an archive file) as indexed_date,
+            # falling back to UTC now (for everything parsed/queued before
+            # update applied to parser). API users use indexed_date to poll for
+            # newly added stories, so a timestamp.  By default, ES stores times
+            # with millisecond granularity.
+            data["indexed_date"] = (
+                content_metadata.get("parsed_date") or datetime.utcnow().isoformat()
```
rahulbot, line 188: Is it more idiomatic in Python to do `content_metadata.get("parsed_date", datetime.utcnow().isoformat())` or what you have here? (honest question to improve my own coding practices)

rahulbot (approved):
Great comments. I left just a few non-blocking questions/comments.

indexer/story_archive_writer.py
philbudne: For the archives written by story_archive_writer using warcio, WARC-Date is always present.
A *BIG* undocumented caveat about the StoryArchiveReader is that it expects (without checking) the WARC file to be one that was written by StoryArchiveWriter, that is, alternating WARC "response" and "metadata" records, with the "metadata" of Content-Type `application/x.mediacloud-indexer+json`, that is, it is NOT a general purpose WARC file reader!!!  Originally I had called the program that reads and queues the contents of archive files "warc-queuer.py", but realized that (1) the format of the archive files is wholly hidden in story_archive_writer.py (which got the reader added to it), and (2) the reader doesn't even pretend to attempt to handle just any WARC file!  More comments needed?!

indexer/workers/importer.py
philbudne: There is one place I remember writing "get" with a default, but realized:
1. I wanted to be extra paranoid and handle the case where the key exists, but has an empty or null result
2. less importantly, the `or` means the default expression is only evaluated if needed.
I can't speak to pythonicity (I'm not a Python zealot, it's just another tool to me, and I'll wash my mouth with soap if I ever utter the word "unpythonic") but short-circuitry is common idiom in C: `!(ptr = function()) || ptr->dereference)` and shell programming `test -d $DIR || mkdir -p $DIR`

pgulley (approved):

thepsalmist (approved):

PR #223, opened 2024-01-30
fix: generate ES id from mcmetadata unique_url_hash

thepsalmist:
fix: generate ES id from mcmetadata unique_url_hash


kilemensi (approved):

indexer/workers/importer.py
```diff
@@ -193,8 +193,7 @@ def import_story(
         """
         response = None
         if data:
-            url = str(data.get("url"))
-            url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
+            url_hash = unique_url_hash(str(data.get("url")))
```
philbudne, line 196: 1. Need to use `normalized_url` from data.  That's the reason we need to reindex.
2. If the value of the `normalized_url` key is anything other than a non-empty string, I think we should quarantine the story, rather than attempting to store using id unique_url_hash("None") or unique_url_hash("").  I realize that these are "should not happen" cases, but it's better to know about them than hide them

kilemensi: I also thought about it as well but when I checked `unique_url_hash` implementation, it does [call](https://github.com/mediacloud/metadata-lib/blob/b67c651550e5a68fa1531b7699c96178f2dd663a/mcmetadata/urls.py#L330) `normalize_url`. Are the two distinct normalization?

philbudne:
> I also thought about it as well but when I checked `unique_url_hash` implementation, it does [call](https://github.com/mediacloud/metadata-lib/blob/b67c651550e5a68fa1531b7699c96178f2dd663a/mcmetadata/urls.py#L330) `normalize_url`. Are the two distinct normalization?

Oh! I hadn't looked inside the implementation of unique_url_hash!

NOTE: @rahulbot

The potential oddness is that if there is a version skew when a
message is handled by parser.py with one version of mcmetadata, which
sets the ContentMetadata "normalized_url", but the message is replayed
from an archive, or sits around in the importer input queue (both will
happen this week), and the importer.py that then picks up the queued
story has a different version of mcmetadata (as it will, because
mcmetadata has been updated, I think, at the very least to discard TCP
ports in URLs), the generated hash used to index the story may not
match the "normalized_url" stored in ES.

I suppose it's a feature that importing does the right/latest thing,
but it means at the best of times we're doing both the normalize and
hash ops twice.

PR #227, opened 2024-02-01
result of falling into importer counter rabbit hole.

philbudne:
In the process of checking out odd stats (deleted documents in ES). I ended up falling into a deep rabbit hole.  Tried calling "create" instead of "index".

Many of the changes are paranoia, and are explicit handling (counting, logging, quarantine) of improbable situations.

Using StoryWorker.incr_stories guarantees all stories are logged with status and URL.

If nothing else (and we don't merge this), it's worth reading both versions of the code as an audit.

### DIFF
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -1,12 +1,12 @@
 """
 elasticsearch import pipeline worker
 """
+
 import argparse
 import logging
 from datetime import datetime
-from typing import Any, Optional, Union
+from typing import Optional, Union
 
-from elastic_transport import ObjectApiResponse
 from elasticsearch.exceptions import ConflictError, RequestError
 from mcmetadata.urls import unique_url_hash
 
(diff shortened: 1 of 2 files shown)


indexer/workers/importer.py
```diff
@@ -40,87 +40,141 @@ def process_args(self) -> None:
...
+            # than pass an empty string, or turn None into "None"
+            self.incr_stories("no-url", "none")
+            raise QuarantineException("no-url")
+        url_hash = unique_url_hash(url)
+
+        # NOTE: indexed_date passed in by process_story
+        try:
+            # logs HTTP op with index name and ID str.
+            # create: raises exception if a duplicate.
+            response = self.elasticsearch_client().create(
```
kilemensi, line 142: Good catch!

kilemensi (approved):
🚀

thepsalmist (approved):

rahulbot (approved):
Looks good to me, but I don't know these codepaths well.
(Happy to chime in when requested to like this, and overall I think things outside user/requirements, integration touchpoints, or other things I'm focused on don't need me to review since I don't want to be a blocker.)

pgulley (approved):

philbudne:
@rahulb -- yes, you might lack familiarity with the code, but your
knowledge of the intended source, semantics and use of the metadata
fields is definitive!

PR #237, opened 2024-02-08
Limit text_content to Lucene term byte length

kilemensi:
### Description

In Elasticsearch, for `text` field to be indexable (i.e. `keyword`), it's length must not exceed the underlying Lucene’s term byte-length limit of **32766**. This PR truncates the `text_content` field if and when it exceeds this limit.

### DIFF
--- a/indexer/tests/test_importer.py
+++ b/indexer/tests/test_importer.py
@@ -6,7 +6,7 @@ from typing import Any, Mapping, Optional, Union
 import pytest
 from elasticsearch import ConflictError, Elasticsearch
 
-from indexer.workers.importer import ElasticsearchImporter
+from indexer.workers.importer import ElasticsearchImporter, truncate_str
 
 
 @pytest.fixture(scope="class", autouse=True)
@@ -96,6 +96,16 @@ test_data: Mapping[str, Optional[Union[str, bool]]] = {
     "text_content": "Lorem ipsum dolor sit amet",
     "indexed_date": datetime.now().isoformat(),
 }
+
+
+def test_truncate_str() -> None:
+    s = "e\u0301"  # é
+    assert truncate_str(s, 0) == ""
+    assert truncate_str(s, 1) == ""
+    assert truncate_str(s, 2) == "é"
+    # Without normalizing to "NFC", e and \u0301 are kept separate
+    assert truncate_str(s, 1, normalize=False) == "e"
+    assert truncate_str(s, 2, normalize=False) == "é"
 
 
 class TestElasticsearchConf:
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -42,13 +42,13 @@ def truncate_str(
     encoded as utf-8.
     """
     if src:
-        src_bytes = src.encode(encoding="utf-8", errors="ignore")
+        src_bytes = src.encode(encoding="utf-8", errors="replace")
     if not src or len(src_bytes) <= max_length:
         return src
     if normalize:
         n_src = unicodedata.normalize("NFC", src)
-        src_bytes = n_src.encode(encoding="utf-8", errors="ignore")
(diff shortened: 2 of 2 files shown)


kilemensi:
Open questions:

1. Do we want to normalize to NFC to increase the chance of truncating at the right code point?
2. Do we want to ignore or replace on errors?

thepsalmist (approved):

philbudne:
> Do we want to normalize to NFC to increase the chance of truncating at the right code point?
Whoa!  No idea!  I had previously been utterly unaware of the Unicode normal forms rabbit hole!!

philbudne:
kilemensi wrote:

> 2. Do we want to ignore or replace on errors?

Since we're starting with a valid str (unicode sequence), PRESUMABLY
the only possible error could be the last unicode character, so ignore
would SEEM to be safe, but nonetheless, unnerving to me!

I'm also uncomfortable with using str(value) too much.

If mypy is complaining a value could be either a string or None,
str(None) would silently yield "None".  I'd rather have explicit
checks (either an assert, which will eventually land the story in
quarantine, or an explicit check for isinstance(x, str) and raising
QuarantineException.

I know it makes the code less simple and pretty, but real life isn't
always either, and I would prefer that we code defensively.  If some
required value is always there, nothing happens, and if, for some
reason it isn't, I'd rather we know about it, and have the buildup of
quarantined story alert us to the problem.

Speaking of which, I had set up a grafana alert at
https://stats.tarbell.mediacloud.org/d/hN1OxidSk/story-indexer-alerts?orgId=1
for when the total quarantined messages (across all workers) exceeds
100K, and it looks like right now we're at 98,588 (and has grown by an
average of 1605 in the past week, so I expect it to trigger in the
next day!)

kilemensi:
>so ignore
would SEEM to be safe, but nonetheless, unnerving to me!
>

Cool, switched to `replace` as the error handling scheme.

>If mypy is complaining a value could be either a string or None,
str(None) would silently yield "None".  I'd ...
>

From the definition of `data`,
```python
        data: dict[str, Optional[Union[str, bool]]] = {}
```
mypy was complaining about the possibility of `bool` rather than `None` in the `text_content` (`truncate_str` can handle None just fine). But, makes sense, I've switched to strict `str` checking and quarantining the story if `text_content` isn't a valid `str`.

kilemensi:
PS: When there is no `url` in `data` @philbudne , should we pass [`"no-url"`](https://github.com/mediacloud/story-indexer/blob/cf805587053acd9b0f351c4666e66cac0c4ba6d9/indexer/workers/importer.py#L78) or [`"none"`](https://github.com/mediacloud/story-indexer/blob/cf805587053acd9b0f351c4666e66cac0c4ba6d9/indexer/workers/importer.py#L134) to `inc_stories`? We seem to use both.

philbudne (approved):

pgulley (approved):

philbudne:
> PS: When there is no `url` in `data` @philbudne , should we pass [`"no-url"`](https://github.com/mediacloud/story-indexer/blob/cf805587053acd9b0f351c4666e66cac0c4ba6d9/indexer/workers/importer.py#L78) or [`"none"`](https://github.com/mediacloud/story-indexer/blob/cf805587053acd9b0f351c4666e66cac0c4ba6d9/indexer/workers/importer.py#L134) to `inc_stories`? We seem to use both.

The second arg is used for logging (I want to be able to grep log
files for a URL to trace its progress through the pipeline), so if a
URL is available (is non-empty), it should be logged.  If there isn't
a URL I guess we should be consistent, but it's unimportant in practice.

In a very few cases I MIGHT have originally chosen to pass something
else, but at a certain point I decided the name of the argument was
"url".  I suppose it could take an optional third arg (in cases where
we want to log the bad data as well as the url) but that seems like a
fine point.

PR #239, opened 2024-02-14
Updates for production historical file processing

philbudne:
* Increase fetcher and parser container counts.
* Add --cleanup option to re-start orphaned files
* Use SQLite3 for local tracking DB: easier to view, back up and modify
* Handle CSV date/times without fractional seconds

### DIFF
--- a/indexer/tracker.py
+++ b/indexer/tracker.py
@@ -6,12 +6,12 @@ Would prefer something replicated/durable (S3, AWS SimpleDB, ES)
 see https://github.com/mediacloud/story-indexer/issues/203
 """
 
-import dbm
 import logging
 import os
+import sqlite3
 import time
 from enum import Enum
-from typing import Any, Callable, Type, cast
+from typing import Any, Optional, Tuple, Type
 
 from indexer.app import AppException
 from indexer.path import DATAROOT
@@ -19,12 +19,27 @@ from indexer.path import DATAROOT
 # basename assumes URLs and local paths both use "/":
 assert os.path.sep == "/"
 
+OLD_SECONDS = 24 * 60 * 60  # age before eligible for cleanup
+
 logger = logging.getLogger(__name__)
 
 
-class TrackerException(AppException):
+class _TrackerException(AppException):
+    """
+    base class for FileTracker exceptions
+    """
+
+
+class TrackerStatus(_TrackerException):
     """
     thrown when file is in some state other than NOT_STARTED
+    """
+
+
+class _TrackerRetry(_TrackerException):
+    """
+    INTERNAL: thrown when a (local file based) tracker is locked
+    and should be retried
     """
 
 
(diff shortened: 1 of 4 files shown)


indexer/tracker.py
```diff
@@ -6,28 +6,43 @@
...
+class _TrackerException(AppException):
+    """
+    base class for FileTracker exceptions
+    """
+
+
+class TrackerStatus(_TrackerException):
     """
     thrown when file is in some state other than NOT_STARTED
     """
```
kilemensi, line 36: May be `StatusError`? TrackerStatus doesn't sound like an exception.

indexer/tracker.py
```diff
@@ -6,28 +6,43 @@
...
 
 
+class _TrackerRetry(_TrackerException):
```
kilemensi, line 39: `_TrackerLocked`? Choice of retrying or giving up is up to the tracker itself I guess.

indexer/tracker.py
```diff
@@ -72,98 +88,223 @@ def basename(self, fname: str) -> str:
...
+
+    def _start_if_available(self) -> None:
+        try:
+            self._retry_open_and_lock()
+            status, ts = self._db_get_status_time()
+            logger.debug("_start_if_available %s %d", status, ts)
+            if status == FileStatus.NOT_STARTED.name or (
+                self.cleanup
+                and status == FileStatus.STARTED.name
+                and (time.time() - ts) > OLD_SECONDS
```
kilemensi, line 152: 🤔 ... so we **won't** delete files that are less than 1 day old even if with `--cleanup`?

indexer/tracker.py
```diff
@@ -72,98 +88,223 @@ def basename(self, fname: str) -> str:
...
+        GDBM and SQLite3 can both raise exceptions when DB locked
+        """
+        for sec in (0.0625, 0.125, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 16.0):
+            try:
+                self._db_open_and_lock()
+                return
+            except _TrackerRetry:
+                logger.debug("_retry_open_and_lock sleeping %g", sec)
+                time.sleep(sec)
+        raise TrackerStatus("locked")
```
kilemensi, line 179: Nitty but isn't `TrackerStatus` tied to the "internal" state of a file i.e. FileStatus (NOT-STARTED, STATED and FINISHED)?

"external" state could be anything filesystem-related; file is locked, non-existing/invalid db path used, etc.

kilemensi (approved):
🚀

pgulley (approved):

thepsalmist (approved):

indexer/tracker.py
philbudne: Replaced with `TrackerNotStartable`

indexer/tracker.py
philbudne: replaced with `TrackerLocked` (exposed to user, see below)

indexer/tracker.py
philbudne: It won't try to re-run files that were started (but not finished) until a day after they were marked as started

indexer/tracker.py
philbudne: Will now (re)raise TrackerLocked to user when file cannot be locked and written to.

Issue #235, opened 2024-02-07
evaluate new ES-ILM backup / redundancy strategy


rahulbot:
With the ILM ES index architecture, @philbudne raised a question about reconsidering our redundancy approach. We now know that we can restore 2-3 months from WARC files in ~2 days. What if we roll-over via ILM to a new index every 2 months, and immediately backup the rolled-over index off-site. Then if we crash restoration is 2-ish days of downloading indexes and recreating the latest (un-backed up) index from .WARC files. I think this is an acceptable downtime, and we can always later add some kind of "hot" duplicate of the latest index if we want. The task here is to consider how to design and implementation for this, whether it would really work, and to make sure it is a good idea.

Related to #157, #231, #54.

rahulbot:
Notes from mtg: sounds doable via an API call (to test), perhaps use 90 days as rough time limit, validate how easy/hard it is to restore an archived index, double-check shard size spec, make sure changes for this don't require re-indexing

thepsalmist:
We can do the ILM policy update via the ILM put API `PUT _ilm/policy/<mc_ILM_policy_id>`. 

When the policy is updated, changes what take effect on our current index `mc_search-0001`, so this means changes would take effect after `mc_search-0002`. So we'll have to let the current policy rollover as per the current ILM rollover definitions as documented [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/update-lifecycle-policy.html)

As per this image, our current max shard size is 17.5GB, so we should anticipate rollover when we ingest about triple our current data (this should be sooner than the alternate rollover action of 365 days)

![Screenshot 2024-02-21 at 17 02 25](https://github.com/mediacloud/story-indexer/assets/49587182/5a1ed59d-624a-4e43-a2fc-c83af2bbdc3d)

rahulbot:
My proposal for setting up backups is:
1. after the first ILM rollover (early-March?) we take an image of that newly-frozen index and back it up off-site (S3?)
2. in March we work on automating that process so the next rollover automatically takes a binary snapshot of the index and backs it up off-site

Closing this and I'll set up different issues accordingly to capture these two tasks to be done at different times.

This supports a two-prong an overall strategy for catastrophic index failure recovery:
* restore prior frozen indexes from off-site storage
* restore "hot" index from warc files (I guess it is OK if a WARC file overlaps a little with latest "cold" index because we'll have URL-based deduplication working)

PR #250, opened 2024-02-26
Fix/update ilm policy

thepsalmist:
Update ILM rollover policy to 90 days, to apply in indices preceding `mc_search-00001` as described here #235


pgulley (approved):

conf/elasticsearch/templates/create_ilm_policy.json
```diff
@@ -13,7 +13,7 @@
       }
     },
     "_meta": {
-      "description": "ILM policy using the hot phase with a rollover of 50gb on max_primary_shard_size or 365 days, whichever comes first",
+      "description": "ILM policy using the hot phase with a rollover of 50gb on max_primary_shard_size or 90 days, whichever comes first",
       "managed": true
     }
```
kilemensi, line 18: Been a long day so ignore me if I'm reading the code wrong but: shouldn't `_meta` be inside `policy` as shown [here](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/ilm-put-lifecycle.html#ilm-put-lifecycle-example)? As is, I don't think we're [storing](https://github.com/mediacloud/story-indexer/blob/42b9e3b1c06a3b0b237f22ead7034294390159f7/indexer/scripts/elastic-conf.py#L131-L133) it in ES.

kilemensi (approved):
🚀

docker/deploy.sh
```diff
@@ -295,7 +295,7 @@ prod)
     # ES index settings are static, prod settings should not change
     ELASTICSEARCH_SHARD_COUNT=30
```
philbudne, line 296: Is it safe to keep the same shard count?  We'll be generating 4x as many per year (every 3 months instead of every 12)?
ISTR something about scaling the number of shards based on server heap memory size, and that each shard requires a thread for searching (and too many shards can exhaust the thread pool), and suggestions about 50GB being the max shard size, but also about smaller goal sizes....  Maybe here: https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html

Seems like this would be a good place to describe how the numbers were selected!

thepsalmist: Seeing the actual production data now, I think reducing the number of shards/index should be safer & help avoiding inefficiencies from oversharding. The initial numbers of 30 shards was from estimation of storage data of upto 12TB over a 5 year period as in the [doc here](https://docs.google.com/document/d/179NElNrlJQX4YVqe4mkSk1yFYgwxBdTy09fyKNsm64w/edit).
As described [here](https://www.elastic.co/elasticon/conf/2016/sf/quantitative-cluster-sizing), this is something that could always change down the line based on the perfomance stats we get in production data.

philbudne (approved):

Issue #243, opened 2024-02-14
delete unnecessary ES index fields


rahulbot:
While working on #229 I revisited the ES index schema, which I believe is at [`conf/elasticsearch/templates/create_index_template.json`](https://github.com/mediacloud/story-indexer/blob/main/conf/elasticsearch/templates/create_index_template.json). If that is accurate, I have some questions about the fields:
* `normalized_article_title`: I think this is used in rss-fetcher for deduplication. Do we use it in story-indexer for the same? If not, why are we storing it? It is not useful to users (it is "lossy").
* `normalized_url`: This is used to compute the hash `_id`, but is it useful for any other reasons? Any reason we are storing it? Users don't search by this and it is similarly "lossy".
* `text_extraction`: What does this field hold? I can't tell from a quick read of the code.
* `text_extraction_method`: This holds the name of the library we used to extract the text from HTML. Is this useful for researchers enough to support storing?

philbudne:
An example metadata record from a WARC file (no `text_extraction` value in any of 451865 WARC records examined).
```
{
  "rss_entry": {
    "link": "https://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
    "title": "The Basic Principles Of would or could",
    "domain": "ttblogs.com",
    "pub_date": "Mon, 19 Feb 2024 04:51:05 -0000",
    "fetch_date": "2024-02-18"
  },
  "http_metadata": {
    "response_code": 200,
    "fetch_timestamp": 1708387428.433925,
    "final_url": "https://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
    "encoding": "utf-8"
  },
  "content_metadata": {
    "original_url": "https://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
    "url": "https://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
    "normalized_url": "http://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
    "canonical_domain": "ttblogs.com",
    "publication_date": "2024-02-19",
    "language": "en",
    "full_language": "en",
    "text_extraction_method": "trafilatura",
    "article_title": "The Basic Principles Of would or could",
    "normalized_article_title": "the basic principles of would or could",
[...]

PR #252, opened 2024-02-27
Fix/update es mapping

thepsalmist:
Update elasticsearch mapping to rm non-required fields, as per issue #243.
The PUT req to configure elasticsearch should update the template. Note: This applies to indices created after `mc-search-00001`
Made an update to get elasticsearch keys from the index template (single source of truth)


pgulley (approved):

indexer/workers/importer.py
```diff
@@ -77,7 +77,14 @@ def incr_pub_date(self, status: str) -> None:
 
     # create once, read-only (tuple)
     # could extract valid keys from index template (schema)??
-    KEYS_TO_SKIP = ("is_homepage", "is_shortened", "parsed_date")
+    KEYS_TO_SKIP = (
```
philbudne, line 80: Given the expanded list, I wonder if it should change to "KEYS_TO_COPY" (less likely to cause pain if mcmetdata changes)?  Or perhaps extract that list from the template?

indexer/workers/importer.py
```diff
@@ -77,7 +77,16 @@ def incr_pub_date(self, status: str) -> None:
 
     # create once, read-only (tuple)
     # could extract valid keys from index template (schema)??
-    KEYS_TO_SKIP = ("is_homepage", "is_shortened", "parsed_date")
+
+    KEYS_TO_COPY = (
```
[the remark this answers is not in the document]
thepsalmist: The keys really are the same, but not sure @kilemensi on how we'll do that without duplicate the concerns in `elastic-conf`, like passing in the file path as an arg & reading the template.json file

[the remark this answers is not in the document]
kilemensi: ... did we design `ElasticMixin` to be reusable across various Elastic workers @thepsalmist ? With a bit of thinking, I'd need to duplicate any concerns.

indexer/elastic.py
```diff
@@ -44,3 +52,23 @@ def elasticsearch_client(self) -> Elasticsearch:
...
         # Connects immediately, performs failover and retries
         return Elasticsearch(hosts.split(","))
+
+    def read_file(self, template_name: str) -> Union[dict, Any]:
+        assert self.args
+        elasticsearch_config_dir = self.args.elasticsearch_config_dir
+        file_path = os.path.join(elasticsearch_config_dir, template_name)
+        with open(file_path, "r") as file:
+            data = file.read()
+        return json.loads(data)
```
kilemensi, line 62:
```suggestion
    def _load_template(self, name: str) -> dict | Any:
        file_path = os.path.join(self.elasticsearch_config_dir, name)
        with open(file_path, "r") as file:
            data = file.read()
        return json.loads(data)
```

1. read_file is too generic
2. I'm assuming this method is supposed to be private-ish
3. Unless there are other benefits, new code should [prefer](https://peps.python.org/pep-0604/) `|` over `Union`

indexer/elastic.py
```diff
@@ -30,6 +32,12 @@ def define_options(self, ap: argparse.ArgumentParser) -> None:
             default=os.environ.get("ELASTICSEARCH_HOSTS") or "",
             help="comma separated list of ES server URLs",
         )
+        ap.add_argument(
+            "--elasticsearch-config-dir",
+            dest="elasticsearch_config_dir",
+            default=os.environ.get("ELASTICSEARCH_CONFIG_DIR") or "",
+            help="ES config files dir",
+        )
```
kilemensi, line 40: define process_args method that will set `self.elasticsearch-hosts` and `elasticsearch-config-dir`. Otherwise, every method that wants to access these variables will have to check `args` first.

indexer/elastic.py
```diff
@@ -44,3 +52,23 @@ def elasticsearch_client(self) -> Elasticsearch:
...
+        assert self.args
+        elasticsearch_config_dir = self.args.elasticsearch_config_dir
+        file_path = os.path.join(elasticsearch_config_dir, template_name)
+        with open(file_path, "r") as file:
+            data = file.read()
+        return json.loads(data)
+
+    def index_template_data(self) -> Any:
+        json_data = self.read_file(template_name="create_index_template.json")
+        return json_data
```
kilemensi, line 66:
```suggestion
    def load_index_template(self) -> Any:
        return self._load_template("create_index_template.json")
```

Since `"create_index_template.json`" is expected to never change, you should define it as a constant

indexer/workers/importer.py
```diff
@@ -75,29 +75,20 @@ def incr_pub_date(self, status: str) -> None:
...
         Import story into Elasticsearch
         """
         content_metadata = story.content_metadata()
         data: dict[str, Optional[Union[str, bool]]] = {}
-
+        # extract valid keys from index template (schema)
+        index_template_data = self.index_template_data()
+        index_template_keys = index_template_data["template"]["mappings"][
+            "properties"
+        ].keys()
```
kilemensi, line 88: You don't have to recreate these keys on every story. Create self.keys_to_copy once at the end of `process_args` function i.e. when all args have been created.

philbudne (approved):

philbudne:
Thought/question: have you tested querying across indices with different schemas to make sure it isn't an issue (as it was with solr)?

pgulley (approved):

indexer/workers/importer.py
```diff
@@ -69,38 +70,42 @@ def process_args(self) -> None:
 
         self.output_msgs = self.args.output_msgs
 
+        index_template = self.load_index_template()
+        if not index_template:
+            logger.error("Elasticsearch Index template not loaded")
+            sys.exit(1)
```
kilemensi, line 76: Lets be consistent... If mixing template is error log + exist, lets do it also in `elastic-conf`.

indexer/tests/test_importer.py
```diff
@@ -84,15 +74,12 @@ def elasticsearch_client() -> Any:
 
 test_data: Mapping[str, Optional[Union[str, bool]]] = {
     "original_url": "http://example.com",
-    "normalized_url": "http://example.com",
     "url": "http://example.com",
     "canonical_domain": "example.com",
     "publication_date": "2023-06-27",
     "language": "en",
     "full_language": "English",
-    "text_extraction": "Lorem ipsum",
```
kilemensi, line 93: Rather than cleaning this, it makes more sense to have a test that ensures the importer is smart enough to only pick the fields it needs and ignore the rest.

kilemensi (approved):

indexer/elastic.py
```diff
@@ -30,17 +32,44 @@ def define_options(self, ap: argparse.ArgumentParser) -> None:
...
             sys.exit(1)
 
         # Connects immediately, performs failover and retries
-        return Elasticsearch(hosts.split(","))
+        return Elasticsearch(self.elasticsearch_hosts.split(","))
+
+    def _load_template(self, name: str) -> dict | Any:
+        if not self.elasticsearch_config_dir:
+            logger.fatal("need --elasticsearch_config_dir or ELASTICSEARCH_CONFIG_DIR")
+            sys.exit(1)
```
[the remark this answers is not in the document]
philbudne: `ElasticMixin` is used by elastic-stats.py, so perhaps put it in an `ElasticConfMixin`? -- If you run a script with `--help` you should only see options that are applicable/used by that program.

If every user of ELASTICSEARCH_CONFIG_DIR also connects to the server, then subclassing `ElasticMixin` is reasonable, otherwise make it an independent mixin??

indexer/elastic.py
```diff
@@ -44,3 +52,23 @@ def elasticsearch_client(self) -> Elasticsearch:
 
         # Connects immediately, performs failover and retries
         return Elasticsearch(hosts.split(","))
+
+    def read_file(self, template_name: str) -> Union[dict, Any]:
+        assert self.args
+        elasticsearch_config_dir = self.args.elasticsearch_config_dir
+        file_path = os.path.join(elasticsearch_config_dir, template_name)
+        with open(file_path, "r") as file:
+            data = file.read()
```
philbudne, line 65: Since this will appear as a method on the app, I think something more specific, like `read_elastic_config` would be a better choice for the name.

Not really an issue in this case, since the files are small, but this reads the entire file into memory AND constructs the JSON tree in memory.  Calling `json.load(file)` puts the business of reading the file into the json library's hands.

indexer/elastic.py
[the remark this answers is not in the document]
kilemensi: I like the idea of introducing `ElasticConfMixin` as subclass of `ElasticMixin` that is to be used by `ElasticConf` and `ElasticsearchImporter` since they both need to connect to the ES server **_and_** load conf/templates.

thepsalmist:
> Thought/question: have you tested querying across indices with different schemas to make sure it isn't an issue (as it was with solr)?

Tested this without an issue, the workaround in such a scenario would be to specify the fields to return in the query using _"source": ["field1","field2"]`

indexer/tests/test_importer.py
thepsalmist: #257

PR #256, opened 2024-02-29
Fix `truncate_str`

kilemensi:
### Description

Current implementation of `truncate_str` tries to replace `decode` errors after truncating a string. This occasionally leads to a final string that is longer than the maximum required length.

This PR ignores `decode` errors because by this point of function execution, the only possible source of `decode` errors is due to the truncation, and by design, we want any bytes after the truncation point _ignored_.

**Aside**
This PR also adds a counter, `truncated`, for stories that have been truncated.

### DIFF
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -48,7 +48,8 @@ def truncate_str(
     if normalize:
         n_src = unicodedata.normalize("NFC", src)
         src_bytes = n_src.encode(encoding="utf-8", errors="replace")
-    return src_bytes[:max_length].decode(encoding="utf-8", errors="replace")
+    # any errors on decode are due to the truncation and should ignored
+    return src_bytes[:max_length].decode(encoding="utf-8", errors="ignore")
 
 
 class ElasticsearchImporter(ElasticMixin, StoryWorker):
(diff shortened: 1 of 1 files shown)


indexer/workers/importer.py
```diff
@@ -155,6 +156,8 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
             raise QuarantineException("no-text")
 
         data["text_content"] = truncate_str(text_content)
+        if len(data["text_content"]) < len(text_content):
+            self.incr_stories("truncated", url)
```
philbudne, line 160: I had been thinking it would be nice to have a counter, so THANK YOU!!

but the reason I hadn't suggested it earlier is that `import_story` always increments the `stories` counter (double counting stories that are truncated), alternatives:
1. a separate counter for story length (two sub-counters, truncated and whole?)
2. a histogram (we don't have any, so there is a question of how to represent it), plus the question of bin size (k, 4k, 16k?)
3. other???

philbudne (changes requested):

indexer/workers/importer.py
kilemensi: _What did I just get myself into?!!_

Good catch. As this counter is more of "a nice to have", option 1 sounds good enough.

thepsalmist (approved):
LGTM

philbudne (approved):

pgulley (approved):

Issue #242, opened 2024-02-14
will ILM rollover make unique-key check fail?


rahulbot:
One of our layers of deduplication relies in `_id` collisions on the ES index, because that is a hash of the normalized URL. Pre-ILM we were using publication date to pick the date to insert into so we didn't need to worry about uniqueness across indexes (assuming pub date extraction is repeatable feels safe enough). Now with ILM we might be inserting a story into the "live" index, but could have fetched it a year ago and it'd be in a different index... which is a problem, no?

How can we retain this useful layer of deduplication with ILM in mind?

rahulbot:
Note: make sure insert is done via index, so this is mostly a task to validate that the duplicate insert fails when it is already in any of the indexes in the alias

kilemensi:
Possibly related: https://github.com/elastic/elasticsearch/issues/44794, https://github.com/elastic/elasticsearch/issues/56992

philbudne:
Yes, 44794 looks like our situation exactly.  It would have been LESS of a problem with the previous scheme (stories routed to indices based on published_date), but if the story was updated such that: the year changed, our extracted date changed (due to either the story or the code changing) it could still have bitten us.

With lookup by _id (in all or all "closed" indices) then create, there's a race (if the rollover happens between the two ops), but even I (an inveterate pessimist) tend to think that's a small risk.  I would like to think our "normal" ingest rate is low enough so that the performance penalty isn't going to be terrible, and I don't know how to get a realistic idea of what the lookup overhead would be....  I suppose it's another variable to consider in how quickly we might want to create new indices....

rahulbot:
Sounds like next step is to test out doing an index check before importing. Kind of critical because we're approaching the shard size limit that would trigger a rollover.

Issue #254, opened 2024-02-28
automate backing up of "warm" ES indexes created by ILM


rahulbot:
Split off from #235. We want to automate the off-site backup of cold indexes right after ILM creates them. It sounds like there are some API methods that could help us do this.

Issue #253, opened 2024-02-28
manually back up first "warm" index after first ILM rollover


rahulbot:
Split from #235: the task here is to manually capture an image of the newly created first "warm" index after ILM does a rollover, set up a place to save these on S3, an put it there.

kilemensi:
I'm assuming we're talking about `"warm"` tier here i.e _read-only_ version of `"hot"` tier but equal in every other way.

If I remember correctly, it was decided we'll never need `"cold"` or `"frozen"` [tiers](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-tiers.html).

rahulbot:
Yes - thanks for the correction 👍🏽

thepsalmist:
As partly mentioned in #257 and as per the discussions on #205 we settled with having all our current ES nodes as `data_hot`. So that we'll be snapshotting indexes that have been rolled_over [by filtering indices](https://www.elastic.co/guide/en/elasticsearch/reference/current/create-snapshot-api.html#create-snapshot-api-example) by `is_write_index` is `false`. All our indices are in the `hot` phase including after rollover but at any given time we [have one write index](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-rollover.html) identified by `is_write_index` is `true`

thepsalmist:
Validated taking manual snapshots on dev as follows

1. register S3 Repository, & assign ES IAM user access to the S3 repo

`
POST  _/snapshot/<snapshot_repository_name>
{
    "type": "s3",
    "settings": {
        "bucket": "s3-repo-name",
        "region": "region",
        "base_path": "base-s3"
    }
}
` 

2.  Create a snapshot - we can specify the index or multiple indices (Our case will be snapshotting indices that have rolled over, i.e `is_write_index=True`), e.g

`POST /_snapshot/<snapshot_repository>/<snapshot_name>
{
    "indices": "mc_search-000006",
    "ignore_unavailable": true,
    "include_global_state": false,
    "metadata": {
        "taken_by": "mc_indexer",
        "taken_because": "test_manual_backup"
    }
}
`

3. We have also validated that we can restore the snapshots in S3 using the Restore Snapshot API

`
PUT _/snapshot/<snapshot_repo_name>/<snapshot_name
{
    "indices": "mc_search-000005",
    "ignore_unavailable": true,
    "include_global_state": false,
    "rename_pattern": "mc_search-(.+)",
[...]

Issue #205, opened 2024-01-08
rearchitect indexes to support archiving?


rahulbot:
For discussion (in connection to #198 and #168):
Via @phil on slack, who asks good questions about the decision to architect indexes by publication year (#67). An alternative would be to freeze indexes after some time, or at some certain size (story count or KB), so we're constantly appending to the latest index, and all prior ones are frozen and backed up as stacked snapshots. 

A quick glance at [ES lifecycle management](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html) suggests there _might_ be built in support for managing this automatically.

For discussion.

thepsalmist:
Background context [here ](https://docs.google.com/document/d/1wuHr0AzrI5XaJfSkqCHSmoSJL0H8OlKp2mEJWLyQipI/edit)

1. Unlike in the current setup, the hot-warm architecture should use `hot-node` for ingestion (We would need at least 2 nodes configured as `hot-nodes` for high availability)

2. Search is parallelized across both the hot and warm nodes. Therefore at least 2 nodes would be needed for high availability.

3. Based on Mediacloud’s use case do we foresee a scenario where we rollover some indices to the cold/frozen tiers, bearing in mind that there will always be search. So do you have an opinion on whether we should implement a cold & or frozen node now?

4. Our initial ingestion will have mixed data from both historical years (pre-2023) and current year 2024, and as we progress the indices should have almost linear data based on publication years. This should affect any search/aggregations based on publication dates/year, since this implies searching across all indices rather than a specific index_year as currently exists. 
Context: Assuming our rollover is based on max shard_size of 50GB, when we start re-indexing the first 5-10 indices will contain mixed data (based on publication dates). A query such as `qs=(NBA)%2520-(Basketball)%2C*&start=01-02-2023%2C12-09-2023&end=01-11-2023`
means searching across all the 5-10 indices?

5. Based on Q 1 & 2 above, In addition the current hardware that we have available (Ramos, Bradley & Woodward), do we have extra hardware to support a scenario where we set a minimum of two(2) `hot/content` nodes two(2)` warm nodes, and if needed `cold & frozen` nodes?

rahulbot:
[...]

PR #257, opened 2024-03-01
Fix/ilm handle duplicates

thepsalmist:
This PR introduces document lookup by `id` before indexing to handle duplicates on index rollover.
We're using ILM, with multiple indices mapped to a single alias hence the use of [Search API ](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html)
#242

### DIFF
--- a/indexer/tests/test_importer.py
+++ b/indexer/tests/test_importer.py
@@ -166,19 +166,48 @@ class TestElasticsearchImporter:
     @pytest.fixture
     def importer(self) -> ElasticsearchImporter:
         importer = ElasticsearchImporter("test_importer", "elasticsearch import worker")
+        importer.elasticsearch_hosts = os.environ.get("ELASTICSEARCH_HOSTS")
         return importer
 
     def test_import_story_success(self, importer: ElasticsearchImporter) -> None:
         test_import_data = {**test_data, "url": "http://example_import_story.com"}
         assert importer.import_story(test_import_data)
 
-    def test_import_story_duplicate(
-        self, importer: ElasticsearchImporter, elasticsearch_client: Any
-    ) -> None:
-        test_import_data = {**test_data, "url": "https://www.bernerzeitung.ch/"}
+    def test_import_story_extra_fields(self, importer: ElasticsearchImporter) -> None:
+        test_import_data = {
+            **test_data,
+            "url": "https://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
+            "normalized_article_title": "the basic principles of would or could",
+            "normalized_url": "http://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
+            "text_extraction_method": "trafilatura",
+        }
+        response = importer.import_story(test_import_data)
+        assert response
+        id = response.get("_id")
+        search_response = importer.elasticsearch_client().search(
+            index="test_mc_search",
+            body={
+                "query": {"bool": {"filter": {"term": {"_id": id}}}},
+            },
+        )
+        assert search_response
+        print(search_response["hits"]["hits"])
+        search_response_keys = search_response["hits"]["hits"][0]["_source"]
+        additional_keys = {
+            key: test_import_data[key]
+            for key in test_import_data
+            if key not in test_data
+        }
+        for key in additional_keys:
+            assert (
+                key not in search_response_keys.keys()
+            ), f"Unexpected field '{key}' found in response."
+
+    def test_import_story_duplicate(self, importer: ElasticsearchImporter) -> None:
+        test_import_data = {**test_data, "url": "https://www.edbernerzeitung.ch12/"}
         url = test_import_data["url"]
         assert isinstance(url, str), "url must be a string"
-        search_response = elasticsearch_client.search(
+        search_response = importer.elasticsearch_client().search(
             index="test_mc_search",
             body={
                 "query": {"bool": {"filter": {"term": {"_id": unique_url_hash(url)}}}},
--- a/indexer/tests/test_importer.py
+++ b/indexer/tests/test_importer.py
@@ -191,7 +191,6 @@ class TestElasticsearchImporter:
             },
         )
         assert search_response
-        print(search_response["hits"]["hits"])
         search_response_keys = search_response["hits"]["hits"][0]["_source"]
         additional_keys = {
             key: test_import_data[key]
--- a/indexer/tests/test_importer.py
+++ b/indexer/tests/test_importer.py
@@ -1,10 +1,10 @@
-import hashlib
 import os
 from datetime import datetime
 from typing import Any, Mapping, Optional, Union
 
 import pytest
 from elasticsearch import ConflictError, Elasticsearch
+from mcmetadata.urls import unique_url_hash
 
 from indexer.workers.importer import ElasticsearchImporter, truncate_str
 
@@ -137,7 +137,7 @@ class TestElasticsearchConf:
 class TestElasticsearchConnection:
     def test_index_document(self, elasticsearch_client: Any) -> None:
         index_name_alias = "test_mc_search"
-        test_id = hashlib.sha256(str(test_data.get("url")).encode("utf-8")).hexdigest()
+        test_id = unique_url_hash(str(test_data.get("url")))
         response = elasticsearch_client.create(
             index=index_name_alias, id=test_id, document=test_data
         )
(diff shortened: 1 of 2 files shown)


indexer/workers/importer.py
```diff
@@ -182,6 +183,18 @@ def import_story(
...
+            search_response = self.elasticsearch_client().search(
+                index=INDEX_NAME_ALIAS,
+                body={
+                    "query": {"bool": {"filter": {"term": {"_id": url_hash}}}},
+                    "size": 0,
+                },
+            )
+            if search_response["hits"]["total"]["value"] > 0:
+                self.incr_stories("dups", url)
+                return None  # mypy explicit return
```
kilemensi, line 197:
```suggestion
            response = self.elasticsearch_client().search(
                index=INDEX_NAME_ALIAS,
                body={
                    "query": {
                        "bool": {
                            "filter": {
                                "term": {
                                    "_id": url_hash,
                                    "_tier": "data_warm",
                                }
                            }
                        }
                    },
                    # don't need the actual documents.
                    "size": 0,
                },
            )
            if response["hits"]["total"]["value"] > 0:
                raise ConflictError()
```

After rollover, we only need to search `data_warm` tier. `create` on the client will still check for duplicates in `data_hot` tier.

indexer/workers/importer.py
```diff
@@ -202,10 +215,11 @@ def import_story(
             # (that should never be seen) should be returned as False (do not archive).
             result = response.get("result", "noresult") or "emptyres"
             self.incr_stories(result, url)  # count, logs result, URL
-            return True
+            return response
```
kilemensi, line 218:
```suggestion
            return url_hash # i.e. response.get("_id")
```

indexer/workers/importer.py
```diff
@@ -202,10 +215,11 @@ def import_story(
...
         except ConflictError:
             self.incr_stories("dups", url)
-            return False
+            return None
+        # return False
```
kilemensi, line 222:
```suggestion
```

indexer/tests/test_importer.py
```diff
@@ -166,9 +166,54 @@ class TestElasticsearchImporter:
...
         assert importer.import_story(test_import_data)
+
+    def test_import_story_extra_fields(self, importer: ElasticsearchImporter) -> None:
+        test_import_data = {
+            **test_data,
+            "url": "https://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
+            "normalized_article_title": "the basic principles of would or could",
+            "normalized_url": "http://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
+            "text_extraction_method": "trafilatura",
+        }
```
kilemensi, line 183:
```suggestion
        test_extra_data = {
            "normalized_article_title": "the basic principles of would or could",
            "normalized_url": "http://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
            "text_extraction_method": "trafilatura",
        }
        test_import_data = {
            **test_data,
            **test_extra_data,
            "url": "https://damienafsoe.ttblogs.com/4775282/the-basic-principles-of-would-or-could",
        }
```

We can then use `test_extra_data` in the final asset without having to loop/dictionary comprehension

indexer/tests/test_importer.py
```diff
@@ -166,9 +166,54 @@ class TestElasticsearchImporter:
...
+        response = importer.import_story(test_import_data)
+        assert response
+        id = response.get("_id")
```
kilemensi, line 186:
```suggestion
        id = importer.import_story(test_import_data)
        assert id
```

indexer/tests/test_importer.py
```diff
@@ -166,9 +166,54 @@ class TestElasticsearchImporter:
...
+        )
+        assert search_response["hits"]["hits"], "No document in search response"
+
+        search_response_keys = search_response["hits"]["hits"][0]["_source"]
+        additional_keys = {
+            key: test_import_data[key]
+            for key in test_import_data
+            if key not in test_data
+        }
+        for key in additional_keys:
```
kilemensi, line 201:
```suggestion
        for key in test_extra_data:
```

indexer/tests/test_importer.py
```diff
@@ -166,9 +166,54 @@ class TestElasticsearchImporter:
...
+        url = test_import_data["url"]
+        assert isinstance(url, str), "url must be a string"
+        search_response = importer.elasticsearch_client().search(
+            index="test_mc_search",
+            body={
+                "query": {"bool": {"filter": {"term": {"_id": unique_url_hash(url)}}}},
+                "size": 0,
+            },
+        )
+        assert search_response["hits"]["total"]["value"] == 0, "Document already exists"
```
kilemensi, line 217: If I'm reading the test correctly, the document should **NOT** exist at this point.

kilemensi (commented):
👍🏽 

---

Do we have any data that can tell us the expected performance hit?

indexer/workers/importer.py
```diff
@@ -168,7 +169,7 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
     def import_story(
         self,
         data: dict[str, Optional[Union[str, bool]]],
-    ) -> bool:
+    ) -> ObjectApiResponse | None:
```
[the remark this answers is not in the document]
philbudne: I agree: avoid returning complex types from a library when a simple/useful one will do.

On the point of whether to start using `|` in place of Union in new code: This seems like a slippery slope, and that it will require immense willpower to avoid making changes in nearby lines.  There's also the question of whether to use `X | None` in place of `Optional[X]`.

Frankly, I find it more disturbing to see a mixture of conventions than to stick with an old one.

Python is constantly mutating, and this won't be the last time a question like this comes up!!!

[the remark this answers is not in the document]
philbudne: I'm reposting the above to slack, a better place for discussion!

indexer/workers/importer.py
thepsalmist: @kilemensi searching by data tier makes sense, however we concluded to have all three hot nodes(ramos,bradley,woodward) based on [this discussions](https://github.com/mediacloud/story-indexer/issues/205)

indexer/workers/importer.py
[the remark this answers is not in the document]
thepsalmist: Based on

> I think standardizing on a runtime version is great (3.10?), and recognize that can be different from decisions about code conventions and consistency (which could be pegged to 3.9 style).

indexer/workers/importer.py
kilemensi: Cool @thepsalmist. If this was agreed, the the backup/snapshot plan also needs to be updated. Otherwise stories such as #253 and #254 still refer to ideas of cold/warm tiers.

thepsalmist: Updated #253 @kilemensi

thepsalmist: >  # don't need the actual documents.

We dont need to return the actual documents,'optimizing' our search query. Query checks for existence

indexer/workers/importer.py
```diff
@@ -168,7 +168,7 @@ def process_story(self, sender: StorySender, story: BaseStory) -> None:
     def import_story(
         self,
         data: dict[str, Optional[Union[str, bool]]],
-    ) -> bool:
+    ) -> str | None:
```
philbudne, line 171: 1. The return value is still only used (in importer.py) as a bool, does the return type need to change?
2. if it does, update the docstring below!!!
3. My preference is to stick with Optional.. It hurts my brain to see Optional one line above (the slippery slope)

indexer/workers/importer.py
```diff
@@ -182,6 +182,18 @@ def import_story(
...
+            # We want to avoid indexing duplicates on ILM index rollover
+            search_response = self.elasticsearch_client().search(
+                index=INDEX_NAME_ALIAS,
+                body={
+                    "query": {"bool": {"filter": {"term": {"_id": url_hash}}}},
+                    "size": 0,
+                },
+            )
+            if search_response["hits"]["total"]["value"] > 0:
+                self.incr_stories("dups", url)
```
philbudne, line 195: I'd rather have a different counter name (dups2?), since it's a different condition.

philbudne (changes requested):

indexer/workers/importer.py
thepsalmist: Cool, I recall there was a limit to how many characters the counter name would be, could gi wuth `ilm-dups` for more clarity

indexer/workers/importer.py
thepsalmist: 1. We're returning the str(url_hash) for tests that involve search hence Optional[str]

kilemensi (approved):
LGTM!

philbudne (approved):

pgulley (approved):

PR #262, opened 2024-03-07
Truncate long `article_title`

kilemensi:
### Description

It seems like some `article_title`s also exceed Lucene's term byte-length limit of 32766. This PR handles them the same way `text_content` was handled.

`article_title` being an optional field (story won't be reject if not present), the counter for title length will have 3 possible labels:

- `empty`: Missing altogether,
- `intact`: Within the allowed length limit, and
-  `trunc`: Exceeded the limit and hence truncated.

### DIFF
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -7,7 +7,7 @@ import logging
 import sys
 import unicodedata
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


philbudne (approved):

thepsalmist (approved):
LGTM

PR #267, opened 2024-03-15
Ft/es snapshots

thepsalmist:
This PR implements taking Elasticsearch snapshots to S3. Implementation uses ES's Snapshot Lifecycle Management (SLM) API  to automate snapshot creation every 14 days.
Snapshots created are incremental and therefore should allow us have a point-in-time restore point for every 2 weeks of ES data.

PS: Removed the dependency of ILM 90-day rollover.


kilemensi (commented):
> ... API to automate snapshot creation every 91 days (Rollover should be every 90 days)

My immediate thoughts/questions (may not be accurate):
1. Are we 100% sure that the two policies we be schedule at _the exact same time_? Otherwise we can't guarantee that 91st day (for snapshot) will be a day after the 90th day (for rollover).
2. The two schedules even if they start at the same time, will drift over time:
> Rollover: 90th day, 180th day, ... 900, 990, ...
> Snapshot: 91st day, 182th day, ... 910, 1001, ...
>
I think you want them to happen every 90 days but one should start a day earlier/later

3. Are the snapshots incremental?

thepsalmist:
> > ... API to automate snapshot creation every 91 days (Rollover should be every 90 days)
> 
> My immediate thoughts/questions (may not be accurate):
> 
> 1. Are we 100% sure that the two policies we be schedule at _the exact same time_? Otherwise we can't guarantee that 91st day (for snapshot) will be a day after the 90th day (for rollover).
> 2. The two schedules even if they start at the same time, will drift over time:
> 
> > Rollover: 90th day, 180th day, ... 900, 990, ...
> > Snapshot: 91st day, 182th day, ... 910, 1001, ...
> 
> I think you want them to happen every 90 days but one should start a day earlier/later
> 
> 3. Are the snapshots incremental?

Yes the snapshots are incremental

Yes, the scheduling is a little bit tricky. Since the count starts on the day of deployment to prod. These would actually overlap since initial rollover would have happened by then. Looking at this!

philbudne:
Seeing this comment:
> Are we 100% sure that the two policies will be schedule at the exact same time? Otherwise we can't guarantee that 91st day (for snapshot) will be a day after the 90th day (for rollover).

I had thought there was some way to trigger actions (like snapshotting) as part of ILM, and that scheduling wouldn't be an issue...

thepsalmist:
> Seeing this comment:
> 
> > Are we 100% sure that the two policies will be schedule at the exact same time? Otherwise we can't guarantee that 91st day (for snapshot) will be a day after the 90th day (for rollover).
> 
> I had thought there was some way to trigger actions (like snapshotting) as part of ILM, and that scheduling wouldn't be an issue...

Yes Ideally, had we gone with the full ILM phase rollovers from `hot->warm->cold/frozen` snapshotting would automatically be done once we rollover int the `cold/frozen` phase.
So right now with all our indices in the `hot` phase, we're trying to achieve this using [Snapshot Lifecycle Management (SLM) APIs ](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/slm-api-put-policy.html)

kilemensi:
@thepsalmist Can you add a small description (in this PR or docs) what this change mean vs what we were aiming for with the 90-day rollover?

thepsalmist:
> @thepsalmist Can you add a small description (in this PR or docs) what this change mean vs what we were aiming for with the 90-day rollover?

Resolved

conf/elasticsearch/templates/create_slm_policy.json
```diff
@@ -0,0 +1,14 @@
+{
```
kilemensi, line 1: I think lets give this policy file a more descriptive name that we can use as policy-id e.g. `biweekly-snapshot-policy.json`, `hot-tier-snapshot-policy.json`, `main-snapshot-policy.json`, etc.

conf/elasticsearch/templates/create_slm_policy.json
```diff
@@ -0,0 +1,14 @@
...
+    "schedule": "0 30 1 */14 * ?",
```
kilemensi, line 2: Minor (and it all depends on what everyone else thinks) but I'd rather have guaranteed dates e.g. 1st and 15th of every month, rather than just _every 14 days_. Depending on the month, we may end up with 3 snapshots in the same month.

conf/elasticsearch/templates/create_slm_policy.json
```diff
@@ -0,0 +1,14 @@
...
+    "name": "<snapshot-{now/d}>",
+    "repository": "$repository_name",
```
kilemensi, line 4:
```suggestion
    "repository": "$repository",
```

conf/elasticsearch/templates/create_slm_policy.json
```diff
@@ -0,0 +1,14 @@
...
+    "config": {
+        "indices": ["mc_search-*"],
+        "ignore_unavailable": true,
+        "include_global_state": true
+    },
+    "retention": {
+        "min_count": 5,
+        "max_count": 10
```
kilemensi, line 12: Roughly how much would 5 - 10 snapshots cost us at the current index sizes?

indexer/scripts/elastic-conf.py
```diff
@@ -47,6 +48,13 @@ def define_options(self, ap: argparse.ArgumentParser) -> None:
             default=os.environ.get("ELASTICSEARCH_ILM_MAX_SHARD_SIZE") or "",
             help="ES ILM policy max shard size",
         )
+        # SLM
+        ap.add_argument(
+            "--slm-repo",
+            dest="elasticsearch_snapshot_repo",
+            default=os.environ.get("ELASTICSEARCH_SNAPSHOT_REPO") or "",
```
kilemensi, line 55: I think lets pick one approach and stick to it. If `repo` is specifically for SLM:
```python
            "--slm-repo",
            dest="slm_repo",
            default=os.environ.get("ELASTICSEARCH_SLM_REPO") or "",
```
i.e. similar to how we've defined ILM variables. On the other hand, if we don't want to tie repo to SLM:
```python
            "--snapshot-repo",
            dest="snapshot_repo",
            default=os.environ.get("ELASTICSEARCH_SNAPSHOT_REPO") or "",
```

indexer/scripts/elastic-conf.py
```diff
@@ -144,6 +154,35 @@ def create_initial_index(self, es: Elasticsearch) -> Any:
                 logger.error("Failed to create Index. Response:%s", response)
             return acknowledged
 
+    def create_slm_policy(self, es: Elasticsearch) -> Any:
+        _TODAY = date.today().strftime("%Y.%m.%d")
```
kilemensi, line 158:
```suggestion
        today = date.today().strftime("%Y.%m.%d")
```
Why caps?

indexer/scripts/elastic-conf.py
```diff
@@ -144,6 +154,35 @@ def create_initial_index(self, es: Elasticsearch) -> Any:
...
+        policy_id = f"slm-policy-{_TODAY}"
```
kilemensi, line 159:
```suggestion
        policy_id = "name of the file"
```

I think `policy_id` needs to be a constant. Otherwise, running it on different days will end creating multiple policies (and activate all of them?).

indexer/scripts/elastic-conf.py
```diff
@@ -144,6 +154,35 @@ def create_initial_index(self, es: Elasticsearch) -> Any:
...
+        repository_name = self.elasticsearch_snapshot_repo
```
kilemensi, line 160:
```suggestion
        repository = self.elasticsearch_snapshot_repo
```

kilemensi (commented):
👍🏽

conf/elasticsearch/templates/create_slm_policy.json
thepsalmist: Testing from the snapshot created on 25/March vs 18/Apr, the bucket size hasn't increased significantly in size, initial 1.5TB now 1.5TB. 
We could see how this changes after 2022 data re-ingestion

doc/elasticsearch/elasticsearch_snapshot.md
```diff
@@ -86,3 +86,62 @@ To create the snapshot without SLM policy using Elasticsearch snapshot API
 ```sh
 curl X -POST "http://localhost:9200/_snapshot/mc_ec_s3_repository/snapshot_{now/d}?wait_for_completion=true"
 ```
+
+### SLM (Snapshot Lifecycle Manager)
+
+We create snapshots using Elasticsearch's Snapshot API, the policy defined [here](../../conf/elasticsearch/templates/create_slm_policy.json)
```
kilemensi, line 92:
```suggestion
We create snapshots using Elasticsearch's Snapshot API, the policy defined [here](../../conf/elasticsearch/templates/bi_weekly_slm_policy.json)
```

doc/elasticsearch/elasticsearch_snapshot.md
```diff
@@ -86,3 +86,62 @@ To create the snapshot without SLM policy using Elasticsearch snapshot API
...
+
+1. Get the list of available snapshots
+
+```
+curl -x GET "http://localhost:9200/_snapshot
+```
+
+2. Restore an index
+
+If we're restoring data to a pre-existing cluster, we can use either of the following methods.
```
kilemensi, line 119: Should we/do you expect us to need this? Isn't it (always?) better to restore to a new setup?

indexer/elastic.py
```diff
@@ -111,3 +111,11 @@ def load_initial_index_template(self) -> Any:
...
         """
         return self._load_template("create_initial_index.json")
+
+    def load_slm_policy_template(self) -> Any:
+        """
+        Load the initial SLM Policy from a JSON file.
+        Returns:
+            Any: SLM policy data.
+        """
+        return self._load_template("bi_weekly_slm_policy.json")
```
kilemensi, line 121: Since `elastic-conf` needs the policy_id as well, I think it's best to define it there and pass it here.
```suggestion
    def load_slm_policy_template(self, policy_id) -> Any:
        """
        Load the initial SLM Policy from a JSON file.
        Args:
            policy_id: 
        Returns:
            Any: SLM policy data.
        """
        return self._load_template(f"{policy_id}_policy.json")
```

indexer/scripts/elastic-conf.py
```diff
@@ -144,6 +154,34 @@ def create_initial_index(self, es: Elasticsearch) -> Any:
                 logger.error("Failed to create Index. Response:%s", response)
             return acknowledged
 
+    def create_slm_policy(self, es: Elasticsearch) -> Any:
+        policy_id = "bi_weekly_slm"
+        repository = self.es_snapshot_repo
+        json_data = self.load_slm_policy_template()
```
kilemensi, line 160: If we define:
```python
CURRENT_POLICY_ID = "bi_weekly_snapshots"
```
at the top of this file, then:
```suggestion
        repository = self.es_snapshot_repo
        json_data = self.load_slm_policy_template(CURRENT_POLICY_ID)
```

kilemensi (approved):
🚀 

---

After starting/activating the SLM, may be we should check that there is only **one** actively running policy just in case.

pgulley (approved):

philbudne (approved):

PR #282, opened 2024-05-07
Added sentry dsn to the news-search-api deployment in docker compose.

pgulley:
Necessary for the news-search-api to initialize the sentry sdk, to get traces to their service.

### DIFF
--- a/docker/docker-compose.yml.j2
+++ b/docker/docker-compose.yml.j2
@@ -140,6 +140,7 @@ x-worker-service-settings: &worker-service-settings
 x-news-search-service-settings: &news-search-service-settings
   environment: &news-search-environment-vars
     <<: *base-environment-vars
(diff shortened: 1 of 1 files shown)


philbudne:
I'd held back on adding SENTRY_DSN to the story-indexer deployment of news-search-api, since there's ongoing work to deploy NSA on its own, and it seemed likely that it would want/need/get its own DSN, or possibly use the mcweb DSN (if needed for distributed tracing)

pgulley:
I was wondering about that- I guess it comes down to a priority question, which really comes down to rahul. If we want to start instrumenting things or looking at performance business before that independent deployment comes down the pipeline, is there any reason not to pop this in there now? @rahulbot?

pgulley:
I'll close this PR then!

PR #292, opened 2024-05-26
Shuffle stories for all queuers that feed queue fetcher

philbudne:
Moved shuffling from rss-puller into new ShufflingStoryProducer class so everything that feeds into queue fetcher benefits.

This should improve fetch rate for "csv" and "rss" (explicit RSS input files) pipelines for 2022 backfill.

### DIFF
--- a/indexer/storyapp.py
+++ b/indexer/storyapp.py
@@ -17,7 +17,7 @@ import random
 import sys
 import threading
 import time
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, TypedDict
 from urllib.parse import urlsplit
 
 from mcmetadata.urls import NON_NEWS_DOMAINS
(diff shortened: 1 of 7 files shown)


indexer/workers/fetcher/rss-puller.py
```diff
@@ -247,6 +238,7 @@ def api_get_and_queue(
                 rss.fetch_date = s.get("fetched_at")
             self.send_story(story)
         # end for s in stories:
+        self.flush_shuffle_batch()
```
kilemensi, line 241: Checking if I got this right; If `_batch_size` is `10` and we got `11` stories:
1. Once we hit `10` (but still in `for s in stories` loop) the underlying sender (assuming `RabbitMQStorySender`) will shuffle and flush stories.
2. When we get out of the `for s in stories`, the underlying sender will still have a single story in `_batch`.

We then force the sending of this one remaining story via `flush_shuffle_batch`.

If the above is correct, minor but isn't it better to continue sleeping/checking for new messages in the `main_loop` first? We can then call `quit` once (which calls `flush_shuffle_batch`) when we know we're done pulling messages?

indexer/storyapp.py
```diff
@@ -210,6 +246,13 @@ def send_story(
             self._channel, story.dump(), exchange, routing_key, props
         )
 
+    def flush(self) -> None:
```
kilemensi, line 249: May be we should add a short doc that this message **MUST** be called (when batching is enabled) before exiting; otherwise there is a real danger of some batched stories remaining in `self._batch` _indefinitely_ if we don't reach `self._batch_size`.

kilemensi (approved):
🚀

PR #294, opened 2024-05-31
update csv queuer APP_BLOBSTORE

thepsalmist:
update csv queuer APP_BLOBSTORE

### DIFF
--- a/indexer/workers/fetcher/csv-queuer.py
+++ b/indexer/workers/fetcher/csv-queuer.py
@@ -21,8 +21,8 @@ DATE_RE = re.compile(r"(\d\d\d\d)[_-](\d\d)[_-](\d\d)")
 
 
 class CSVQueuer(Queuer):
-    AWS_PREFIX = "HIST"  # S3 env var prefix
     HANDLE_GZIP = True  # just in case
+    APP_BLOBSTORE = "CSV"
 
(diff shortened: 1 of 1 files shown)


indexer/workers/fetcher/csv-queuer.py
```diff
@@ -21,8 +21,8 @@
 
 
 class CSVQueuer(Queuer):
-    AWS_PREFIX = "HIST"  # S3 env var prefix
     HANDLE_GZIP = True  # just in case
+    APP_BLOBSTORE = "CSV"
```
philbudne, line 25: Looking at the diff, I see you correctly realized (and I didn't) that APP_BLOBSTORE replaced AWS_PREFIX (since it now applies to BackBlaze as well).

BUT, I think keeping the value "HIST" would make sense, since we're unlikely to want to introduce new CSV_S3_XXX environment variables to configure the keys, and it was already working using the "HIST" (or "QUEUER") key(s).

philbudne (changes requested):

indexer/workers/fetcher/csv-queuer.py
philbudne: Nevermind the above comment: the keys actually being used are the QUEUER_XX_XXXX keys, so having the APP name be CSV is fine

philbudne (approved):

PR #295, opened 2024-06-04
Elasticsearch register s3 and fs repository 

thepsalmist:
This PR implements registration and verification of a repository s3/fs during elasticsearch setup. Default to filesystem (fs) for snapshots in ES running docker containers.


indexer/scripts/elastic-conf.py
```diff
@@ -162,6 +225,16 @@ def create_initial_index(self, es: Elasticsearch) -> Any:
     def create_slm_policy(self, es: Elasticsearch) -> Any:
         CURRENT_POLICY_ID = "bi_weekly_slm"
         repository = self.es_snapshot_repo
+
+        if self.es_snapshot_type == "fs":
+            self.register_fs_repository(es)
+        elif self.es_snapshot_type == "s3":
+            self.register_s3_repository(es)
+        # To Add Backblaze support
```
philbudne, line 233: Backblaze provides an S3 compatible API, I found this:

https://github.com/elastic/elasticsearch/issues/21283#issuecomment-828002399

> B2 has S3 compatible API. It works fine for us. We are using a snapshot like this:
```
{
  "type": "s3",
  "settings": {
    "bucket": "elastic-backup",
    "region": "",
    "endpoint": "s3.us-west-001.backblazeb2.com"
  }
}
```

docker/deploy.sh
```diff
@@ -328,6 +330,7 @@ prod)
     ELASTICSEARCH_ILM_MAX_AGE="90d"
     ELASTICSEARCH_ILM_MAX_SHARD_SIZE="50gb"
     ELASTICSEARCH_HOSTS=http://ramos.angwin:9200,http://woodward.angwin:9200,http://bradley.angwin:9200
+    ELASTICSEARCH_SNAPSHOT_TYPE="s3"
```
kilemensi, line 333: 1. Either **unset** `fs` based settings like `_LOCATION` or/and add a warning in the register method if both `_LOCATION` and `_BUCKET` are set.
2. Based on Phil's finding, add a `_ENDPOINT` [setting](https://www.elastic.co/guide/en/elasticsearch/reference/8.14/repository-s3.html#repository-s3-client) when using `s3` to support pointing to B2 instead of S3 if/when we want to.
3. What does staging use?

docker/deploy.sh
```diff
@@ -635,7 +638,10 @@ add ELASTICSEARCH_CONFIG_DIR
 add ELASTICSEARCH_CONTAINERS int
 add ELASTICSEARCH_HOSTS
 add ELASTICSEARCH_SNAPSHOT_CRONJOB_ENABLE # NOT bool!
+add ELASTICSEARCH_SNAPSHOT_TYPE
 add ELASTICSEARCH_SNAPSHOT_REPO allow-empty
+add ELASTICSEARCH_SNAPSHOT_SETTINGS_LOCATION
+add ELASTICSEARCH_SNAPSHOT_SETTINGS_BUCKET #private
```
kilemensi, line 644: Given that we already have `ELASTICSEARCH_SNAPSHOT_REPO` and seeing all these snapshot settings in one place like this, I think we should use `ELASTICSEARCH_SNAPSHOT_REPO` prefix for all of them:
```sh
add ELASTICSEARCH_SNAPSHOT_REPO
add ELASTICSEARCH_SNAPSHOT_REPO_TYPE
add ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_BUCKET
add ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_ENDPOINT
```

This would nicely match the expected JSON:
```json
{
  "type": "s3",
  "settings": {
    "bucket": "my-bucket",
    "endpoint": "s3.amazonaws.com"
  }
}
```

with `ELASTICSEARCH_SNAPSHOT_REPO` being the name.

docker/docker-compose.yml.j2
```diff
@@ -87,7 +87,7 @@ x-es-service-settings: &es-service-settings
     ES_JAVA_OPTS: "-Xms1g -Xmx1g"
     discovery.type: single-node
 {% endif %}
-    path.repo: "/var/backups/elasticsearch"
+    path.repo: {{elasticsearch_snapshot_settings_location}}
```
kilemensi, line 90: This works (I think) but I'd prefer to have this point to a [parent directory](https://www.elastic.co/guide/en/elasticsearch/reference/8.14/snapshots-filesystem-repository.html#snapshots-filesystem-repository) where one can have one or more actual snapshot repos in. For example, if:
```yml
  path.repo: "/var/backups/elasticsearch" # or anywhere easily accessible to dev e.g. elasticsearch folder in the current repo
```
then `ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_LOCATION` can just be a relative path like `mc-backup` and Elasticsearch will resolve it to `/var/backups/elasticsearch/mc-backup`

indexer/scripts/elastic-conf.py
```diff
@@ -48,12 +48,30 @@ def define_options(self, ap: argparse.ArgumentParser) -> None:
             help="ES ILM policy max shard size",
         )
         # SLM
+        ap.add_argument(
+            "--es-snapshot-s3-bucket",
+            dest="es_snapshot_s3_bucket",
+            default=os.environ.get("ELASTICSEARCH_SNAPSHOT_SETTINGS_BUCKET") or "",
+            help="ES snapshot S3 bucket",
+        )
```
kilemensi, line 56: May be I'm in the minority but I'd rather pass things like S3 bucket, endpoints or filesystem location as environment variables...

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +123,42 @@ def main_loop(self) -> None:
...
+                settings={"location": self.es_snapshot_fs_location, "compress": True},
+            )
+            if response and response.get("acknowledged", False):
+                logger.info("Filesystem repository registered successfully.")
+            else:
+                logger.error("Failed to register filesystem repository.")
+        else:
+            logger.info("Filesystem repository already exists.")
+
+    def register_s3_repository(self, es: Elasticsearch) -> None:
```
kilemensi, line 148: Looking at `register_s3_repository` and `register_fs_repository`, I think the code is very similar; the only difference being `settings`. I think
1. a simple `if`/`else` that initializes `settings` based on `type` could be all we need. If it's too long, then
2. method(s) that return `settings` based on `type` should work.

indexer/scripts/elastic-conf.py
```diff
@@ -162,6 +225,16 @@ def create_initial_index(self, es: Elasticsearch) -> Any:
     def create_slm_policy(self, es: Elasticsearch) -> Any:
         CURRENT_POLICY_ID = "bi_weekly_slm"
         repository = self.es_snapshot_repo
+
+        if self.es_snapshot_type == "fs":
+            self.register_fs_repository(es)
+        elif self.es_snapshot_type == "s3":
+            self.register_s3_repository(es)
```
kilemensi, line 232: I don't think `create_slm_policy` needs to know or care about snapshot repo `type`... at best, it may only want to check if there is a registered snapshot with a given name or not.

indexer/scripts/elastic-conf.py
```diff
@@ -76,6 +95,14 @@ def process_args(self) -> None:
...
         self.ilm_max_shard_size = self.args.ilm_max_shard_size
         self.es_snapshot_repo = self.args.es_snapshot_repo
+        self.es_snapshot_type = self.args.es_snapshot_type
+        self.es_snapshot_s3_bucket = self.args.es_snapshot_s3_bucket
+        self.es_snapshot_fs_location = self.args.es_snapshot_fs_location
+
+        if not self.es_snapshot_s3_bucket:
+            logger.warning(
+                "--es-snapshot-s3-bucket or ELASTICSEARCH_SNAPSHOT_SETTINGS_BUCKET not set"
+            )
```
kilemensi, line 105: Why a warning? What if I'm using `fs` with location?

kilemensi (commented):
👍🏽 ... almost there

docker/deploy.sh
```diff
@@ -328,6 +330,7 @@ prod)
     ELASTICSEARCH_ILM_MAX_AGE="90d"
     ELASTICSEARCH_ILM_MAX_SHARD_SIZE="50gb"
     ELASTICSEARCH_HOSTS=http://ramos.angwin:9200,http://woodward.angwin:9200,http://bradley.angwin:9200
+    ELASTICSEARCH_SNAPSHOT_REPO_TYPE="s3"
```
kilemensi, line 333: If we don't `unset` `ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_LOCATION`, don't we end up with:
```sh
ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_LOCATION="mc_story_indexer"
ELASTICSEARCH_SNAPSHOT_REPO_TYPE="s3"
```
and yet `ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_BUCKET` will be set from private config?

docker/docker-compose.yml.j2
```diff
@@ -435,6 +435,9 @@ services:
       ELASTICSEARCH_SHARD_COUNT : {{elasticsearch_shard_count}}
       ELASTICSEARCH_SHARD_REPLICAS: {{elasticsearch_shard_replicas}}
       ELASTICSEARCH_SNAPSHOT_REPO: {{elasticsearch_snapshot_repo}}
+      ELASTICSEARCH_SNAPSHOT_SETTINGS_BUCKET: {{elasticsearch_snapshot_settings_bucket}}
+      ELASTICSEARCH_SNAPSHOT_SETTINGS_LOCATION: {{elasticsearch_snapshot_settings_location}}
+      ELASTICSEARCH_SNAPSHOT_TYPE: {{elasticsearch_snapshot_type}}
```
kilemensi, line 440:
```suggestion
      ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_BUCKET: {{elasticsearch_snapshot_repo_settings_bucket}}
      ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_LOCATION: {{elasticsearch_snapshot_repo_settings_location}}
      ELASTICSEARCH_SNAPSHOT_REPO_TYPE: {{elasticsearch_snapshot_repo_type}}
```

indexer/scripts/elastic-conf.py
```diff
@@ -54,6 +66,12 @@ def define_options(self, ap: argparse.ArgumentParser) -> None:
             default=os.environ.get("ELASTICSEARCH_SNAPSHOT_REPO") or "",
             help="ES snapshot repository name",
         )
+        ap.add_argument(
+            "--es-snapshot-type",
+            dest="es_snapshot_type",
+            default=os.environ.get("ELASTICSEARCH_SNAPSHOT_REPO_TYPE") or "fs",
+            help="ES snapshots type, default fs",
+        )
```
kilemensi, line 74:
```suggestion
        ap.add_argument(
            "--es-snapshot-repo-type",
            dest="es_snapshot_repo_type",
            default=os.environ.get("ELASTICSEARCH_SNAPSHOT_REPO_TYPE") or "",
            help="ES snapshots type, default fs",
        )
```

If we have to set a default value, it makes sense to set it in one place only... either in `deploy.sh` or here but not both.

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,50 @@ def main_loop(self) -> None:
...
 
+    def repository_exists(self, es: Elasticsearch, repo_name: str) -> bool:
+        status = False
+        try:
+            response = es.snapshot.get_repository(name=repo_name)
+            if repo_name in response:
+                status = True
+        except Exception as e:
+            logger.error("Error validating repository: %s", e)
+        return status
```
kilemensi, line 127:
```suggestion
    def _repository_exists(self, es: Elasticsearch) -> bool:
        try:
            response = es.snapshot.get_repository(name=self.es_snapshot_repo)
            return repo_name in response
        except Exception as e:
            logger.exception("Error validating repository: %s", name)
        return False
```

1. repo name should be available to all methods in this class (self.`es_snapshot_repo`)
2. [`logger.exception`](https://docs.python.org/3/library/logging.html#logging.Logger.exception) includes error stack trace by default (if you wanted to include `e` in log message)

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,50 @@ def main_loop(self) -> None:
...
+
+    def register_repository(self, es: Elasticsearch, repo_type: str) -> bool:
+        status = False
+        if not self.repository_exists(es, self.es_snapshot_repo):
+            settings = {"location": self.es_snapshot_fs_location, "compress": True}
+            if repo_type == "s3":
+                settings = {
```
kilemensi, line 134: I think it's better to change this to use early return to avoid nesting ifs/try/etc.
```python
def register_repository(self, es: Elasticsearch) -> bool:
        if self._repository_exists(es, self.es_snapshot_repo):
            logger.info("repository already exists: %s", self.es_snapshot_repo)
            return False

        # continue as if repo doesn't exist
```

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,50 @@ def main_loop(self) -> None:
...
+                    "bucket": self.es_snapshot_s3_bucket,
+                    "endpoint": self.es_snapshot_s3_endpoint,
+                }
```
kilemensi, line 137: You need to handle cases where `self.es_snapshot_repo_type` is neither `fs` or `s3` here.

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,50 @@ def main_loop(self) -> None:
...
+            try:
+                response = es.snapshot.create_repository(
+                    name=self.es_snapshot_repo,
+                    type=repo_type,
+                    settings=settings,
+                )
+
+                if response and response.get("acknowledged", False):
+                    logger.info(
+                        f"{repo_type.capitalize()} repository registered successfully."
```
kilemensi, line 148: Don't use f-string for logs.

indexer/scripts/elastic-conf.py
```diff
@@ -162,6 +226,14 @@ def create_initial_index(self, es: Elasticsearch) -> Any:
     def create_slm_policy(self, es: Elasticsearch) -> Any:
         CURRENT_POLICY_ID = "bi_weekly_slm"
         repository = self.es_snapshot_repo
+
+        if not self.register_repository(es, self.es_snapshot_type):
+            logger.error(
+                "Elasticsearch snapshot repository does not exist: %s",
+                self.es_snapshot_type,
+            )
+            sys.exit(1)
```
kilemensi, line 235: Why `exit` here when we return a `bool` everywhere else?

indexer/scripts/elastic-conf.py
thepsalmist: Of the valid ES repo types (azure,gcs,fs,s3,source,url,hdfs), we only need `fs` & `s3`, backblaze is catered for by type `s3`

kilemensi: Yes, you need to rase an error/return `False` if `self.es_snapshot_repo_type` is neither `fs` or `s3`... looking at the code, you're assuming it's always either `fs` or `s3`.

docker/deploy.sh
thepsalmist: Not sure I understand this @kilemensi , when generating the compose file the `ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_BUCKET` should have been sourced from private config & we don't use the `*_SETTINGS_LOCATION` when repo type is `s3`

kilemensi: What I mean is these settings go in pairs @thepsalmist. You either use

```sh
ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_LOCATION="folder"
ELASTICSEARCH_SNAPSHOT_REPO_TYPE="fs"
```
or
```sh
ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_BUCKET="bucket"
ELASTICSEARCH_SNAPSHOT_REPO_TYPE="s3"
```
.

Having `_LOCATION` set if `_TYPE` is `s3` _could_ mean someone wanted to use `_BUCKET` but mistakenly set `_LOCATION` instead or forgot to update `_TYPE` to `s3` i.e. hard to debug runtime errors.

So we should either `unset` in `deploy.sh` depending on `_TYPE` and/or issue a warning/error in `elastic-conf.py` if both `_BUCKET` and `_LOCATION` are set at the same time.

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,55 @@ def main_loop(self) -> None:
...
+        except Exception:
+            logger.exception("Error validating repository: %s", repo_name)
+        return False
+
+    def register_repository(self, es: Elasticsearch) -> bool:
+        status = False
+        repo_type = self.es_snapshot_repo_type
+        if not repo_type:
+            logger.error("Repository type must be either 'fs' or 's3'")
+            return status
```
kilemensi, line 133:
```suggestion
        repo_type = self.es_snapshot_repo_type
        if repo_type not in ["fs", "s3"]:
            logger.error("Unsupported repository type: '%s'. Must be either 'fs' or 's3'", repo_type)
            return False
```

This should make sure _repo_name_ is set **and** it's one of the supported type. Don't think we need _status_.

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,55 @@ def main_loop(self) -> None:
...
+
+        if self.repository_exists(es):
+            logger.info("%s repository already exists.", repo_type)
+            return True
+
+        if repo_type == "s3":
+            settings = {
+                "bucket": self.es_snapshot_s3_bucket,
+                "endpoint": self.es_snapshot_s3_endpoint,
```
kilemensi, line 142: Isn't `endpoint` optional i.e. should only be set if `self.es_snapshot_s3_endpoint` not empty?

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,55 @@ def main_loop(self) -> None:
...
+                name=self.es_snapshot_repo,
+                type=repo_type,
+                settings=settings,
+            )
+
+            if response and response.get("acknowledged", False):
+                logger.info("%s repository registered successfully.", repo_type)
+                status = True
+            else:
+                logger.error("Failed to register %s repository.", repo_type)
```
kilemensi, line 161:
```suggestion
            acknowledged = False
            if response:
                acknowledged = response.get("acknowledged", False)
            if acknowledged:
                logger.info("Successfully registered repository: %s", self.es_snapshot_repo)
            else:
                logger.error("Failed to register repository: %s", self.es_snapshot_repo)
            return acknowledged
```

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,55 @@ def main_loop(self) -> None:
...
+        except Exception:
+            logger.exception(
+                "Exception occurred while registering %s repository", repo_type
+            )
+        return status
```
kilemensi, line 166:
```suggestion
            logger.exception(
                "Failed to register repository: %s", self.es_snapshot_repo
            )
        return False
```

Of course, we can also include `self.es_snapshot_repo_type` in the error message (e.g. `Failed to register repository: fs://mc_story_backups`) but it's important to include name.

indexer/scripts/elastic-conf.py
```diff
@@ -162,6 +231,14 @@ def create_initial_index(self, es: Elasticsearch) -> Any:
     def create_slm_policy(self, es: Elasticsearch) -> Any:
         CURRENT_POLICY_ID = "bi_weekly_slm"
         repository = self.es_snapshot_repo
+
+        if not self.register_repository(es):
+            logger.error(
+                "Elasticsearch snapshot repository does not exist: %s",
+                self.es_snapshot_repo_type,
```
kilemensi, line 238:
```suggestion
                "Elasticsearch snapshot repository does not exist: %s",
                self.es_snapshot_repo,
```

indexer/scripts/elastic-conf.py
philbudne: Two choices:
1. For AWS specify "region" and not "endpoint"; library generates endpoint; For any other provider specify "endpont"
2. Have just "endpoint": for AWS configure https://s3.REGION.amazonaws.whatever

philbudne (approved):
looks reasonable.  See my comment that "endpoint" is sufficient for either AWS S3 or Backblaze

indexer/scripts/elastic-conf.py
```diff
@@ -96,6 +116,59 @@ def main_loop(self) -> None:
...
+
+        if repo_type == "s3":
+            settings = {
+                "bucket": self.es_snapshot_s3_bucket,
+                "endpoint": self.es_snapshot_s3_endpoint,
+            }
+        elif repo_type == "fs":
+            settings = {"location": self.es_snapshot_fs_location}
+        else:
+            logger.error("Unsupported repository type: %s", repo_type)
```
kilemensi, line 149:
```suggestion
        else: # repo_type == "fs"
            settings = {"location": self.es_snapshot_fs_location}
```

We've already checked at the top of the method `if repo_type not in ["fs", "s3"]` so at this point it must be one of the two.

indexer/scripts/elastic-conf.py
kilemensi: I think Option 2 is cool since we may drop `ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_BUCKET` completely... but that we need to document that `endpoint` should include `bucket`

My question was if we're sticking with Option 1 (as that would allow us to easily use all options we may need):
```python
        repo_name = self.es_snapshot_repo
        if repo_type == "s3":
            if not any([self.es_snapshot_s3_bucket, self.es_snapshot_s3_endpoint]):
                logger.error("Failed to register s3 repository %s: bucket or endpoint required, none provided", repo_name)
                return False

            settings = {}
            if self.es_snapshot_s3_bucket:
                settings["bucket"] = self.es_snapshot_s3_bucket
            if self.es_snapshot_s3_endpoint:
                settings["endpoint"] = self.es_snapshot_s3_endpoint
            # Should be able to add other settings such as region, etc.
```

kilemensi (approved):

philbudne:
> I think Option 2 is cool since we may drop `ELASTICSEARCH_SNAPSHOT_REPO_SETTINGS_BUCKET` completely... but that we need to document that `endpoint` should include `bucket`

The endpoint URL as I suggested includes the region, but not the bucket....

PR #303, opened 2024-06-18
Add Airtable Update Script

pgulley:
Added a script which will report new deployments to airtable, for a little more visibility


philbudne (commented):
Questions/thoughts:
1. Since the new script is not (currently) invoked inside a container, can the pyairtable requirement be in the `dev` list?
2. It will require an up-to-date venv (or else it will lose update events), so put a check in at the top (after the test for an active venv)?  Maybe test `if ! python3 -mindexer.scripts.airtable-update --help >/dev/null; then COMPLAIN`?
3. The python script requires that the API key env var is exported.  Rather than putting the export in the config files, I think it would be more sanitary to put `export ENV_VAR` in deploy.sh before invoking the program.
4. Using the `App` class/framework makes the program less portable (can't be moved to another repo), so I'd advise not depending on it...
5. regarding BASE_ID; I agree it seems a bit icky to wire in a "magic number", especially when the program already depends on the API key coming from the private config... BUT I have NO idea how sensitive the BASE_ID is, or under what conditions (if ever) it might change...  It CERTAINLY would change if some other organization wanted to run a copy of the software!!
6. Our usual import of the datetime package is `import datetime as dt` then reference `dt.timestamp` etc...

pgulley:
Good comments, thank you! I think the only thing I'm uncertain of here is whether the venv check I added at line 43 is what you meant...

philbudne (approved):

indexer/scripts/airtable-update.py
```diff
@@ -0,0 +1,91 @@
...
+    # Get hardware ids:
+    hardware_table = api.table(MEAG_BASE_ID, table_name="Hardware - Angwin Cluster")
+
+    hardware_res = [
+        hardware_table.first(formula=match({"name": hardware_name}))
+        for hardware_name in hardware_names
+    ]
+    if None in hardware_res:
+        # The names of the machines in the angwin cluster are all valid options.
+        names = [h["fields"]["name"] for h in hardware_table.all()]
```
kilemensi, line 39: On codebase_res (`L25`), `"Name"` is used instead of `"name"`. Are these fields names case-insensitive?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.12
```
kilemensi, line 2: Minor-ish but `story-indexer` is still on Python `3.10` I believe.

indexer/scripts/airtable-update.py
```diff
@@ -0,0 +1,91 @@
...
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(
+        description="A utility for updating an airtable deployment record"
+    )
+
+    parser.add_argument("--name", help="additional deployment name")
+    parser.add_argument(
+        "--env",
+        choices=["production", "staging", "other"],
```
kilemensi, line 81: Could be wrong but I think `DEPLOY_TYPE` uses `prod` instead of `production` (`SENTRY_ENVIRONMENT` uses `production`)

kilemensi (approved):
🚀 

--

1. By the time `deploy.sh` calls the Airtable script, the actual deployment is already done. However, looking at the Airtable script, it will exist without recording the deployment if, for e.g., provided software or hardware name doesn't match what is in the Airtable. Shouldn't it record the deployment _somewhere_ on the Airtable even in this case so that there is actual full record of all deployments?
2. For complex app like `story-indexer` that can be deployed multiple times for different pipelines, etc., wouldn't it make sense to record at least major options such as `PIPELINE_TYPE`?

pgulley:
Hopefully the final update here, I've updated this to use a single central source for the update script so as to not repeat myself across our many repositories

PR #314, opened 2024-07-05
Chore/setup es blobstore credentials

thepsalmist:
Script to add S3/B2 credentials to Elasticsearch keystore. Needed for snapshot & restore to AWS S3/Bcackblaze

### DIFF
--- a/doc/elasticsearch/elasticsearch_configuration.md
+++ b/doc/elasticsearch/elasticsearch_configuration.md
@@ -93,14 +93,13 @@ The RPM and Debian packages will configure this setting automatically. No furthe
 
 ### Configuring S3 Repository for Snapshot/Restore
 
-We are using S3 for Elasticsearch Snapshot & Restore
+We are using Elasticsearch's S3 Repository plugin for Elasticsearch Snapshot & Restore
 
 The bucket `mediacloud-elasticsearch-snapshots` has been created with the relevant policy to allow ES snapshot access to the S3 bucket.
 
 The AWS secure credntials are added to Elasticsearch's keystore as follows:
 
-    `bin/elasticsearch-keystore add s3.client.default.access_key`
-    `bin/elasticsearch-keystore add s3.client.default.secret_key`
+    `./elasticsearch/blobstore_credentials.sh`
 
 #### Register S3 repository
 
--- a/elasticsearch/blobstore_credentials.sh
+++ b/elasticsearch/blobstore_credentials.sh
@@ -63,7 +63,7 @@ if [ "$BLOBSTORE" != "s3" ] && [ "$BLOBSTORE" != "b2" ]; then
 fi
 
 PRIVATE_CONF_DIR="es-credentials-setup"
-mkdir -p $PRIVATE_CONF_DIR
+run_as_login_user mkdir -p $PRIVATE_CONF_DIR
 chmod go-rwx $PRIVATE_CONF_DIR
 
 cd $PRIVATE_CONF_DIR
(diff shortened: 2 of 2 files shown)


kilemensi:
🤔 ... I think lets first clearly document all the steps we currently have to do to add credentials in all ES nodes.

1. Do we need to create the keystore or is it created as part of installation?,
2. Is it password protected or not (is there a need)?,
3. Are the keystore contents backed-up as part of our backup (should they be?), etc.
4. ???

Once this is done, then we'll be in a better position to answer your question:
>Thinking if/how to handle multiple hosts & if this should leave on code repo??
>

thepsalmist:
> 🤔 ... I think lets first clearly document all the steps we currently have to do to add credentials in all ES nodes.
> 
> 1. Do we need to create the keystore or is it created as part of installation?,
> 2. Is it password protected or not (is there a need)?,
> 3. Are the keystore contents backed-up as part of our backup (should they be?), etc.
> 4. ???
> 
> Once this is done, then we'll be in a better position to answer your question:
> 
> > Thinking if/how to handle multiple hosts & if this should leave on code repo??

1. The keystore is not created automatically as part of installtion, needs to be created (in every ES node) by `bin/elasticsearch-keystore add` https://www.elastic.co/guide/en/elasticsearch/reference/current/elasticsearch-keystore.html
2. Currently not password protected, there's an option to create with password 
3. The `keystore` is not part of the backup done from the Snapshot Lifecycle Manager. Not sure if we need to make modifications to backup the keystore file-path

philbudne:
Noticed this PR and wanted to share some musings/questions (without any specific answers):

I looked at it and thought "what does that have to do with blobstore.py?"  Is there some ES-specific term (like repository or keystore) that could be used instead?

I see it creates a new top level directory for elasticsearch; we've tried to avoid top level clutter, and we already have `docker/elastic-deploy.sh`, which isn't ideal.  I originally proposed that we run RabbitMQ outside docker, and in that case I wouldn't want a rabbitmq top level directory as well as an ES directory.  One way to avoid cognitive dissonance would be to silently think of the docker directory as the deploy directory.  I guess another would be to rename the directory!!  It's a topic worthy of group discussion.

Could the installation of the keys be done in elastic-deploy.sh???  I realize it probably requires waiting for the ES cluster to come up...

As a P.S.  Some time ago, I remember looking and seeing that I thought elastic-deploy.sh added a file to sources.list.d for the ES repo, but that was NOT the state of the currently running servers!

When possible, I think install scripts should be re-runable/idempotent: testing whether a step has been taken or not before taking the step, so that it's a no-op to re-run the script...

thepsalmist:
> Noticed this PR and wanted to share some musings/questions (without any specific answers):
> 
> I looked at it and thought "what does that have to do with blobstore.py?" Is there some ES-specific term (like repository or keystore) that could be used instead?
> 
> I see it creates a new top level directory for elasticsearch; we've tried to avoid top level clutter, and we already have `docker/elastic-deploy.sh`, which isn't ideal. I originally proposed that we run RabbitMQ outside docker, and in that case I wouldn't want a rabbitmq top level directory as well as an ES directory. One way to avoid cognitive dissonance would be to silently think of the docker directory as the deploy directory. I guess another would be to rename the directory!! It's a topic worthy of group discussion.
> 
> Could the installation of the keys be done in elastic-deploy.sh??? I realize it probably requires waiting for the ES cluster to come up...
> 
> As a P.S. Some time ago, I remember looking and seeing that I thought elastic-deploy.sh added a file to sources.list.d for the ES repo, but that was NOT the state of the currently running servers!
> 
> When possible, I think install scripts should be re-runable/idempotent: testing whether a step has been taken or not before taking the step, so that it's a no-op to re-run the script...

Good suggestions from the above,
1. the suggestion of an elasticsearch TLD was from an initial conversation on `elastic-deploy.sh` not being ideal with the docker/ directory. I'm also for the discussion on renaming the docker directory to something along deploy scripts. 

2. Move the script as part of elastic_deploy answers the other questions on naming et.al

PR #327, opened 2024-08-16
Switch to fetching RSS files from B2

philbudne:
Add indexer/workers/fetcher/common.py with RSS_FETCHER_URL_BASE (for B2)

Use in indexer/workers/fetcher/rss-queuer.py
and indexer/workers/fetcher/rss_utils.py (and solve a mypy typing complaint)


pgulley (approved):

kilemensi (approved):
🚀

PR #332, opened 2024-08-25
indexer/workers/hist-fetcher.py: try to detect non-existent objects

philbudne:
Right now when processing a "historic" CSV file where HTML is to be found by legacy system `downloads_id` in an S3 bucket, if the object doesn't exist:
1. Nothing is logged
2. No counter is incremented
3. The story is retried (in an hour, or more)
4. Retries are repeated until the story is quarantined

This change tries to detect non-existent objects by having hist-fetcher examine the boto `ClientError` exception, logging and counting as "not-found" and discarding the story if the error was "404"

### DIFF
--- a/indexer/workers/hist-fetcher.py
+++ b/indexer/workers/hist-fetcher.py
@@ -17,6 +17,7 @@ import time
 from typing import Any, Dict, Optional
 
 import boto3
+from botocore.exceptions import ClientError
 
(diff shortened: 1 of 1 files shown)


kilemensi (approved):

thepsalmist (approved):
🚀

pgulley (approved):

Issue #340, opened 2024-10-23
Use _reindex api to generate a small test version of our index


pgulley:
While we're testing out different ways our index structure may be improved, we will want the ability to dry-run on real data. 
We should:
1. Run through the steps of generating a test index which contains some random sample of our real data, using the _reindex api- either via kibana, or directly via curl
2. Document that process here
3. Produce a script or other utility for replicating that process with various index mapping choices.

PR #343, opened 2024-10-29
chore: ES reindex implementation

thepsalmist:
This PR addresses Issue #340

### DIFF
--- a/bin/run-elastic-reindex.sh
+++ b/bin/run-elastic-reindex.sh
@@ -11,23 +11,71 @@
 #
 
 # Usage:
-# Modify the variables ES_HOST, SOURCE_INDEX, and DEST_INDEX as needed.
+# Modify the variables ES_HOST, SOURCE_INDEX, and DEST_SUFFIX and MAX_DOCS as needed.
 # Run this script using: sh reindex.sh
-# ./run-elastic-reindex.sh ES_HOST SOURCE_INDEX DEST_INDEX
+# ./run-elastic-reindex.sh @
 #
 
-# Example:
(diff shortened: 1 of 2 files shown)


kilemensi:
My (very) quick thoughts:

1. Won't the script be better in Python (it could be just me and my `sh` allergies)?
2. How do we plan on using this script?
   i. Would we want to use it to `reindex` documents matching a certain query? In that case, the script should support a `query`,
   ii. Do we care about any impact running this script could have on the cluster?, Then the script should support `throttling` or `slicing`, etc.

thepsalmist:
> My (very) quick thoughts:
> 
> 1. Won't the script be better in Python (it could be just me and my `sh` allergies)?
> 2. How do we plan on using this script?
>    i. Would we want to use it to `reindex` documents matching a certain query? In that case, the script should support a `query`,
>    ii. Do we care about any impact running this script could have on the cluster?, Then the script should support `throttling` or `slicing`, etc.

On the support for `throttling`, I'm not sure we wan't to define the `requests_per_second` to set throttling on initial `reindex` request without knowing the impact on the cluster. 
Since we've settled on Kibana being part of the cluster, I believe we can set this using the [Rethrottle API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html#docs-reindex-rethrottle) based on the monitoring stats

m453h (approved):
Looks good 🚀

pgulley (approved):
I trust the details here- thanks for including good documentation!

kilemensi (approved):
My bash-fu ain't all that but LGTM

PR #346, opened 2024-10-30
Set canonical_domain & Third try at using canonical URL from mcmetadata

philbudne:
The second (to try avoid HTML parser slowness, and extract canonical URL up front) performed worse (fewer stories extracted) than original.

This uses XMLPullParser to detect/discard feed documents by looking at the first tag only (no canonical URL extraction, it was too painful having two places that might supply it).

With apples to apples comparison (runs with same CSV of 100K objects, note: staging runs on a random sampling of 50K) created EXACTLY the same number of stories in ES as original version.

### DIFF
--- a/indexer/workers/parser.py
+++ b/indexer/workers/parser.py
@@ -204,7 +204,7 @@ class Parser(StoryWorker):
         check if mcmetadata.extract found a canonical URL,
         validate it, and update Story object.
         """
-        log_url = self._log_url(story)
+        log_url = self._log_url(story)  # before setting final_url!!
 
         canonical_url = cmd.canonical_url
         if not canonical_url or canonical_url == NEED_CANONICAL_URL:
@@ -215,8 +215,6 @@ class Parser(StoryWorker):
         # from a source we filter out!!!
         if not self.check_story_url(canonical_url):
             return False  # logged and counted
-
-        log_url = self._log_url(story)  # before setting final_url!!
 
         with story.http_metadata() as hmd:
             # story_archive_writer wants hmd.final_url
--- a/indexer/workers/parser.py
+++ b/indexer/workers/parser.py
@@ -3,27 +3,29 @@ metadata parser pipeline worker
 """
 
 import datetime as dt
-import io
 import logging
 from collections import Counter
 from typing import Any
+from xml.etree.ElementTree import XMLPullParser
 
 # PyPI:
 import mcmetadata
 from bs4.dammit import UnicodeDammit
 
-# xml.etree.ElementTree.iterparse doesn't have recover argument
-# (fatal error on control characters in element text), so using lxml
-from lxml.etree import iterparse
-
 # local:
 from indexer.app import run
-from indexer.story import NEED_CANONICAL_URL, BaseStory
+from indexer.story import NEED_CANONICAL_URL, BaseStory, ContentMetadata
 from indexer.storyapp import StorySender, StoryWorker
 from indexer.worker import QuarantineException
 
 QUARANTINE_DECODE_ERROR = True  # discard if False
-FEED_TAGS = ("rss", "feed", "rdf", "channel")  # must be tuple for startswith
+
+# must be tuple for endswith
+# ordered (somewhat) by frequency
+FEED_TAGS = ("rss", "feed", "rdf", "channel", "urlset", "sitemapindex")
+
+# omit total so stackable in grafana, no fetching done by mcmetadata
+SKIP_EXTRACT_STATS_ITEMS = {"total", "fetch"}
 
 logger = logging.getLogger("parser")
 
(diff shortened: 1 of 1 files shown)


philbudne:
Run against WARC input file of stories that all previously caused massive delays:
```
invoking indexer.workers.parser --test-file-prefix test-warcs/parser-hang --rabbitmq-url x
2024-10-30 17:52:46,737 | INFO | indexer.app | STATSD_URL not set
2024-10-30 17:52:46,737 | INFO | indexer.sentry | SENTRY_DSN not found. Not logging errors to Sentry
2024-10-30 17:52:46,830 | INFO | parser | parsing 3358172997: 9211438 characters
2024-10-30 17:52:46,895 | INFO | indexer.storyapp | feed: 3358172997
2024-10-30 17:52:46,952 | INFO | parser | parsing 3357305750: 9216266 characters
2024-10-30 17:52:47,037 | INFO | indexer.storyapp | feed: 3357305750
2024-10-30 17:52:47,100 | INFO | parser | parsing 3357270854: 9209002 characters
2024-10-30 17:52:47,183 | INFO | indexer.storyapp | feed: 3357270854
2024-10-30 17:52:47,239 | INFO | parser | parsing 3357170719: 9208027 characters
2024-10-30 17:52:47,288 | INFO | indexer.storyapp | feed: 3357170719
2024-10-30 17:52:47,316 | INFO | parser | parsing 3356828136: 7785501 characters
2024-10-30 17:52:47,791 | INFO | indexer.storyapp | feed: 3356828136
2024-10-30 17:52:47,900 | INFO | parser | parsing 3356789661: 7233451 characters
2024-10-30 17:52:47,917 | INFO | indexer.storyapp | feed: 3356789661
2024-10-30 17:52:47,984 | INFO | parser | parsing 3356670306: 7233058 characters
2024-10-30 17:52:47,997 | INFO | indexer.storyapp | feed: 3356670306
2024-10-30 17:52:47,997 | INFO | indexer.storyapp | processed 7 stories
```

indexer/workers/parser.py
```diff
@@ -136,84 +146,118 @@ def _save_metadata(self, story: BaseStory, mdd: dict[str, Any], html: str) -> bo
...
+            self.incr_stories("no-canonical-url", log_url)
+            return False  # logged and counted
 
         # now that we FINALLY have a URL, make sure it isn't
         # from a source we filter out!!!
         if not self.check_story_url(canonical_url):
-            return ""
+            return False  # logged and counted
+
+        log_url = self._log_url(story)  # before setting final_url!!
```
kilemensi, line 219: `log_url` is already declared (unless `self.check_story_url` can cause it to change).

indexer/workers/parser.py
```diff
@@ -136,84 +146,118 @@ def _save_metadata(self, story: BaseStory, mdd: dict[str, Any], html: str) -> bo
...
-                        og_url = element.get("content")
-        # end for event: prefer canonical_url
-        canonical_url = canonical_url or og_url
-        # XXX need html.unescape??  seems not?
-        if not canonical_url:
-            self.incr_stories("no-canonical-url", self._log_url(story))
-            return ""
-
-        logger.info("%s: canonical URL %s", story.rss_entry().link, canonical_url)
+        if "<!doctype" in data[0:128].lower():
```
kilemensi, line 177: I'm assuming it's unlikely we'll ever come documents with any other `doctype` besides `html`.

kilemensi (approved):

indexer/workers/parser.py
```diff
@@ -136,84 +146,118 @@ def _save_metadata(self, story: BaseStory, mdd: dict[str, Any], html: str) -> bo
...
+            return True
+
+        try:
+            pp = XMLPullParser(["start"])
+            pp.feed(data)  # pass substring? (first 8K?)
```
m453h, line 182: I could be wrong, but shouldn't we make sure we strip leading whitespaces in the string before parsing it? I think XMLPullParser would always throw an error if there are some whitespaces

m453h (approved):

pgulley (approved):

indexer/workers/parser.py
philbudne: The XML parser throws exceptions on "<!doctype ....>", and none of the feeds that were causing problems have them (I've never seen a feed document with one, tho they are apparently possible in XML).  The RSS world is pretty wild (lots of different things called RSS), but this source says RSS docs should not have a doctype: http://www.xmlbook.info/5-RSS.phtml

This code will only be used for the nov/dec 2021 period, so it will (currently) only see usage from a narrow band of time, and only from sites mediacloud pulled from in that time period.

Yes, the slice is to catch anything with leading whitespace, but I may have seen docs with leading comments "".  It will treat a doctype inside a comment as HTML, but I think it's reasonable to say that's unlikely in a feed document, and it's better to err on the side of thinking it's HTML and trying `extract` on it than to possibly throw away stories unless/until we run up against lots of documents that cause problems.

I suppose searching for "<!doctype html" would be safer, and safest would be to remove the test altogether and let the XML parser throw an exception...

PR #347, opened 2024-11-02
indexer/scripts/elastic-stats.py: report individual index stats again

philbudne:
Currently storing Kibana data in main ES cluster, which messes up grafana displays, so report per-index stats (for indices that don't start with a dot), and only count health counts for non-dot indices.


kilemensi (approved):
🚀 

-----

Couldn't figure out why the method is called `index` though... I thought it'd be named report/record/capture/stats, or something along those lines.

Issue #345, opened 2024-10-30
Process needed for directly editing story entries in the ES archive


pgulley:
We recently indexed a bunch of stories with an "original_url" of "mediacloud.org/need_canonical_url", without correctly setting the canonical domain for those stories, so they're effectively hidden from our front end entirely.  This presents a new problem- can we update a document in place within ES? We need a script which will, for each problematic story:

1. Grab the story id
2. Determine the correct url/canonical domain 
3. Write an update to ES to correct the canonical_domain field

@m453h, I think the first steps is just to duplicate @philbudne's histogram of all of the offending stories, and go from there! 
If I'm understanding the problem correctly, the data is all present, just in the wrong place- so there's no need to go back to the original data that we queued up, right?

philbudne:
My (updated) query script (using raw ES API):
[mc.py.txt](https://github.com/user-attachments/files/17589864/mc.py.txt)

philbudne:
Looks like multiple entries can be updated in one call with `bulk()` API, so multiple entries can be pulled at once, and then be sent back in a single call?
https://elasticsearch-py.readthedocs.io/en/stable/helpers.html

PR #348, opened 2024-11-05
Implement canonical domain update script

m453h:
This PR introduces a script for updating incorrect canonical domains for documents stored in Elasticsearch. 

The implementation of this script uses:

- The [scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html) for the retrieval of documents in batches with configurable batch size. 

- The [Bulk Helpers](https://github.com/mediacloud/story-indexer/issues/345#:~:text=https%3A//elasticsearch%2Dpy.readthedocs.io/en/stable/helpers.html) to perform batch updates across multiple documents simultaneously, significantly reducing the number of requests sent to Elasticsearch.

Example of the usage of the script :
```
./bin/run-elastic-update-canonical-domain.sh --elasticsearch-hosts  http://localhost:9200 --index=mc_search-000001 --buffer=2000 --batch=1000 --query='{                                                                                                                        
    "query": {
        "bool": {
            "must": [
                {
                    "term": {
                        "canonical_domain": "mediacloud.org"
                    }
                }
            ]
        }
    }
}'
```
Addresses #345 

====

### Update
The following improvements have been done following observations from @philbudne :

 - To address concerns regarding the usage of [scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html) the implementation has been updated to use [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) to retrieve documents that should be updated.

- Support for  `query_string` has been added thus the script can also be called as follows:

```
./bin/run-elastic-update-canonical-domain.sh --elasticsearch-hosts  http://localhost:9200 --index=mc_search-000001 --buffer=2000 --batch=1000 --query='canonical_domain:mediacloud.org AND original_url:"http\://mediacloud.org/need_canonical_url"' --type=query_string
```

### DIFF
--- a/indexer/scripts/elastic-canonical-domain-update.py
+++ b/indexer/scripts/elastic-canonical-domain-update.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, Generator, List, Literal, Optional, Tuple
 
 import mcmetadata
 from elasticsearch import Elasticsearch
-from elasticsearch.helpers import bulk, scan
+from elasticsearch.helpers import bulk
 
 from indexer.app import App
 from indexer.elastic import ElasticMixin
@@ -17,6 +17,8 @@ class CanonicalDomainUpdater(ElasticMixin, App):
 
     def __init__(self, process_name: str, descr: str) -> None:
         super().__init__(process_name, descr)
+        self.pit_id: Optional[str] = None
+        self.keep_alive: str = ""
         self.es_client: Optional[Elasticsearch] = None
         self.batch_size: int = 0
         self.updates_buffer: List[Dict[str, Any]] = []
@@ -77,6 +79,13 @@ class CanonicalDomainUpdater(ElasticMixin, App):
             help="The elasticsearch query format (supported values are: [DSL, query_string])",
         )
 
+        ap.add_argument(
+            "--keep_alive",
+            dest="keep_alive",
+            default="1m",
+            help="How long should Elasticsearch keep the PIT alive",
+        )
+
     def process_args(self) -> None:
         """
         Process command line arguments and set instance variables.
@@ -100,6 +109,7 @@ class CanonicalDomainUpdater(ElasticMixin, App):
         self.buffer_size = int(args.buffer_size)
         self.query = args.query
         self.query_format = args.query_format
+        self.keep_alive = args.keep_alive
 
     def initialize(self) -> None:
         """
@@ -112,6 +122,11 @@ class CanonicalDomainUpdater(ElasticMixin, App):
         app.args = parser.parse_args()
         app.process_args()
         self.es_client = self.elasticsearch_client()
+        self.pit_id = self.es_client.open_point_in_time(
+            index=self.index_name,
+            keep_alive=self.keep_alive,
+        ).get("id")
+        logger.info("Successfully opened Point-in-Time with ID %s", self.pit_id)
 
     def build_query(self) -> Tuple[bool, Optional[Dict[str, Any]], Optional[str]]:
         """
@@ -175,7 +190,7 @@ class CanonicalDomainUpdater(ElasticMixin, App):
 
     def get_documents_to_update(self) -> Generator[Dict[str, Any], None, None]:
         """
-        Get documents that need to be updated using scroll API
+        Get documents that need to be updated using search_after
 
         Yields:
             Document dictionaries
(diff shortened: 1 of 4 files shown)


pgulley (approved):

philbudne:
Some thoughts/questions:
1. Would it be possible/practical to take a `query_string` instead of DSL?  It might make it (too?) easy to use?!
2. Regarding the example, there are a _few_ mediacloud.org pages in the index (for whatever reason), so also querying for `original_url` of `http://mediacloud.org/need_canonical_domain` would eliminate them
3. The scroll API page you reference says to avoid using it for more than 10K hits; is this something we can safely ignore??

m453h:
> Some thoughts/questions:
> 
> 1. Would it be possible/practical to take a `query_string` instead of DSL?  It might make it (too?) easy to use?!
> 2. Regarding the example, there are a _few_ mediacloud.org pages in the index (for whatever reason), so also querying for `original_url` of `http://mediacloud.org/need_canonical_domain` would eliminate them
> 3. The scroll API page you reference says to avoid using it for more than 10K hits; is this something we can safely ignore??

Hey @philbudne interesting thoughts and questions:

1. On adding query string: I believe I can modify the script so that we could support both `DSL` and `query_string`, the user would only pass the query_string as:
`original_url:"http://mediacloud.org/need_canonical_url" AND canonical_domain:"blogdomago.com"`
 
    But we would construct the query as:
    ```
    "query": {
           "query_string": {
                "query": "original_url:\"http://mediacloud.org/need_canonical_url\" AND canonical_domain:\"mediacloud.org\""
              }
     }
    ```
With this validation, execution would just remain the same without needing a lot of modification. Let me work on this and push the changes.

2. Regarding the example: The script accepts query as an argument so we can always change the query as we see fit to match the documents that will need to be updated at the time of executing the script.

3. Regarding the use of Scroll API: I did see the warning, regarding scroll API, I did a bit of more research before the implementation in most cases I see the issue with scroll API boils down to maintaining the search context on the server which is costly.I went with Scroll API since:
   - The script would be executed once as a maintenance task, meaning we avoid the risk of having too many search contexts open. Based on the documentation, it states: "Scrolling is not intended for real-time user requests, but rather for processing large amounts of data, such as reindexing the contents of one data stream or index into a new data stream or index with a different configuration."
    -   Scroll API is provided by the [scan helper](https://elasticsearch-py.readthedocs.io/en/7.x/helpers.html#scan) out of the box which seems to best fit this scenario.

  I'm open to making changes and utilize [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after).

indexer/scripts/elastic-canonical-domain-update.py
```diff
@@ -0,0 +1,355 @@
...
+            assert self.es_client
+            count_response = self.es_client.count(index=self.index_name, body=query)
+            count = count_response.get("count")
+            if count is not None:
+                return int(count)
+        except Exception as e:
+            logger.error("Error getting document count: %s", e)
+        return None
+
+    def get_documents_to_update(self) -> Generator[Dict[str, Any], None, None]:
```
kilemensi, line 191: Minor but since we're searching across the network, shouldn't we name this `fetch_xxxx`/`search_xxxx`/etc. to signify this? get_ may imply everything is in memory.

indexer/scripts/elastic-canonical-domain-update.py
```diff
@@ -0,0 +1,355 @@
...
+                "doc": {
+                    "canonical_domain": canonical_domain,
+                },
+            }
+
+            self.updates_buffer.append(update_action)
+
+            # Check if the buffer is full and flush updates
+            if len(self.updates_buffer) >= self.buffer_size:
+                self.flush_updates()
```
kilemensi, line 278: 1. I think `bulk_update` may be a more apt name than `flush_updates` here.
2. Code is very clear, I don't think the comment adds anything to it.

indexer/scripts/elastic-canonical-domain-update.py
```diff
@@ -0,0 +1,355 @@
...
+            success, failed = bulk(
+                client=self.es_client,
+                actions=self.updates_buffer,
+                refresh=False,
+                raise_on_error=False,
+            )
+
+            is_failed_list = isinstance(failed, list)
+            if is_failed_list:
+                assert isinstance(failed, list)
```
kilemensi, line 300: Aren't we doing multiple checks for `isinstance(failed, list)`?

indexer/scripts/elastic-canonical-domain-update.py
```diff
@@ -0,0 +1,355 @@
...
+                assert isinstance(failed, int)
+                failed_count = failed
+            logger.info("Bulk update: %s successful, %s failed", success, failed_count)
+
+            if is_failed_list:
+                assert isinstance(failed, list)
+                for error in failed:
+                    logger.error("Failed to update: %s", error)
+        except Exception as e:
+            logger.error("Bulk update failed: %s", e)
```
kilemensi, line 312: In an exception handler, prefer to use [`logger.exception`](https://docs.python.org/3/library/logging.html#logging.exception) as it includes error info automatically.

indexer/scripts/elastic-canonical-domain-update.py
```diff
@@ -0,0 +1,355 @@
...
+        finally:
+            # Clear the buffer
+            self.updates_buffer = []
+
+    def main(self) -> None:
```
kilemensi, line 317: As we're extending App, overriding `main_loop` may be simpler than `main` as lots of plumbing is already setup for you.

philbudne (approved):

indexer/scripts/elastic-canonical-domain-update.py
```diff
@@ -0,0 +1,355 @@
...
+            help="Elasticsearch query string to filter documents for processing",
+        )
+
+        ap.add_argument(
+            "--format",
+            dest="query_format",
+            default="query_string",
+            choices=["DSL", "query_string"],
+            help="The elasticsearch query format (supported values are: [DSL, query_string])",
+        )
```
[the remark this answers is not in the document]
m453h: @kilemensi , here we are trying to allow the user to either enter the full `DSL` or just the `query_string` part, as per  [this comment](https://github.com/mediacloud/story-indexer/pull/348#issuecomment-2460233560) i.e. someone could just pass the query:

`--query=canonical_domain:mediacloud.org AND original_url:"http\://mediacloud.org/need_canonical_url" `

indexer/scripts/elastic-canonical-domain-update.py
```diff
@@ -0,0 +1,355 @@
...
+            buffer_size (int): Maximum number of update operations to buffer before flushing to Elasticsearch
+            query (str): Elasticsearch query string to filter documents for processing
+
+        Note:
+            Calls the parent class's process_args() first to handle any inherited argument processing.
+        """
+        super().process_args()
+
+        args = self.args
+        assert args
```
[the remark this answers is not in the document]
m453h: This is necessary since mypy can't tell that self.args is not none

kilemensi (approved):

indexer/scripts/elastic-canonical-domain-update.py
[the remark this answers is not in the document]
m453h: Cool, I got you...I think I can make one quick final touch on this before merging 👍

PR #349, opened 2024-11-06
elastic-stats.py: add more gauges, incl gc, thread_pool rejects, circuit breaker trips

philbudne:
Add more gauges:

per node:
nodes.jvm.mem.heap_used_percent
nodes.jvm.gc.collection_count
nodes.jvm.gc.collection_millis
nodes.http.client_count
nodes.thread_pool.POOLNAME.{queue,active,rejected,completed}
nodes.breakers.BREAKERNAME.tripped

cluster health:
cluster.health.status w/ color={green,yellow,red}
cluster.health.shards w/ status={initializing,unassigned,delayed_unassigned}

### DIFF
--- a/indexer/scripts/elastic-stats.py
+++ b/indexer/scripts/elastic-stats.py
@@ -36,35 +36,121 @@ class ElasticStats(ElasticMixin, IntervalMixin, App):
                         logger.debug(" %s %s", path, v2)
                         self.gauge(path, v2)
 
+    def indices_stats(self) -> None:
+        es = self.elasticsearch_client()
+        # see https://github.com/mediacloud/story-indexer/issues/199
+        stats = cast(Dict[str, Any], es.indices.stats())  # fetches /_stats
+
+        # top level keys: "_shards", "_all", "indices"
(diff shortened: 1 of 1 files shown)


indexer/scripts/elastic-stats.py
```diff
@@ -36,35 +36,121 @@ def index(self, name: str, data: StatsDict) -> None:
...
+            self.gauge("indices.health", count, labels=[("color", color)])
+
+    def node_stats(self) -> None:
+        es = self.elasticsearch_client()
+        stats = cast(Dict[str, Any], es.nodes.stats().raw)
+
+        for node_id, node_data in stats["nodes"].items():
+            node_name = node_data["name"].split(".")[0]
+
+            def ng(ni: list[str], value: int) -> None:
```
kilemensi, line 71: Minor but with all possible gotchas with defining functions in a loop, isn't it better to just defined it outside? We're only saving passing in `node_name` if I'm reading the code right..

kilemensi (approved):

pgulley (approved):

PR #354, opened 2024-11-24
indexer/worker.py: Have check_output_queues check "data" disk space.

philbudne:
With stacks running on a single server, the "data" volume is almost certainly on the same disk as rabbitmq queues, so pause queuing if the data volume is too full.

This is to try to keep "historical" processing stacks running on smaller servers from filling the disk if queues back up further down the line (ie; the importer queue when ES is overloaded).

### DIFF
--- a/indexer/worker.py
+++ b/indexer/worker.py
@@ -1015,7 +1015,7 @@ class Producer(QApp):
                 # percentage of blocks available to regular users
                 disk_percent_available = 100 * usage.free / usage.total
                 self.gauge("data-disk-free", disk_percent_available)
-                if disk_percent_available >= min_disk_free:
+                if disk_percent_available < min_disk_free:
                     report_status("disk-space-low")
                     return
                 what = "disk"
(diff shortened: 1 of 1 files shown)


indexer/worker.py
```diff
@@ -972,16 +985,31 @@ def check_output_queues(self) -> None:
...
+                # if queues back up.  If running outside docker,
+                # "data" should have been created to hold queue
+                # tracker files.
+                try:
+                    fs = os.statvfs("data")
+                except FileNotFoundError:
+                    return
+
+                # percentage of blocks available to regular users
+                disk_percent_available = 100 * fs.f_bavail / fs.f_blocks
```
kilemensi, line 1003: Haven't checked the implementation details but this looks like what [`disk_usage`](https://docs.python.org/3/library/shutil.html#shutil.disk_usage) returns?

kilemensi (approved):
🚀

indexer/worker.py
philbudne: My Unix orientation is showing!

Looking at the source for shutil.disk_usage, it uses statvfs, but also works on Windows 🤣 using an NT API

philbudne:
@pgulley can you review?  Should I switch to the Windows friendly API?!

pgulley:
> @pgulley can you review? Should I switch to the Windows friendly API?!

If it's a simple change, let's use the API with the widest coverage. Not that I anticipate running this on a windows server anytime (knock on wood) but it's probably the best OSS practice

pgulley (approved):

PR #356, opened 2024-12-11
Add importer --no-import / deploy -z option

philbudne:
This is to enable "off-line" processing of historical data, direct to the (B2) archive, so that the 2022 (csv) "dip" period can be re-run using blind fetching from S3 with canonical URL extraction before possibly removing the old stories from the index.

Also silences complaints about Dockerfile, stack deploy command & docker-compose.yml

Preserves expanded docker-compose.yml.dump files as docker-compose.yml.save-TAG

### DIFF
--- a/indexer/workers/importer.py
+++ b/indexer/workers/importer.py
@@ -71,12 +71,27 @@ class ElasticsearchImporter(ElasticConfMixin, StoryWorker):
             help="Disable output to archiver",
         )
 
+        # While it's wasteful to run importer processes that don't do
+        # importing, but it's a rare case (created for "offline"
+        # processing of historical data) and this was less invasive
+        # than altering the plumbing of the pipeline.
+        ap.add_argument(
+            "--no-import",
+            action="store_true",
+            default=False,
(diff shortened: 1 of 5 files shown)


indexer/workers/importer.py
```diff
@@ -71,12 +71,27 @@ def define_options(self, ap: argparse.ArgumentParser) -> None:
...
     def process_args(self) -> None:
         super().process_args()
         assert self.args
         logger.info(self.args)
 
         self.output_msgs = self.args.output_msgs
+        self.no_import = self.args.no_import
+
+        if not self.output_msgs and self.no_import:
+            logger.error("no output AND no import?!")
```
kilemensi, line 94: `error` but happy to continue? Shouldn't that be a `warn` if that's the case?

kilemensi (approved):

pgulley (approved):

indexer/workers/importer.py
philbudne: point taken!

PR #360, opened 2024-12-17
Try to suppress noisy messages from metadata libraries

philbudne:
* update to mcmetadata 1.4.0
* remade requirements files
* added App.sentry_init
* call [sentry_]ignore_loggers from Parser.sentry_init()


pgulley (approved):
lgtm

Issue #355, opened 2024-12-04
Bug Report: TypeError: Object of type datetime is not JSON serializable


thepsalmist:
An error occurs when attempting to serialize a datetime object to JSON when passing in rabbitmq_headers as extra_metadata to the `StoryArchiveWriter`

Sample rabbitmq_headers

```
rabbitmq_headers: {

x-death: [
{"count":"110L","exchange":"''","original-expiration":"'120000'","queue":"'fetcher-fast'","reason":"'expired'","routing-keys":"['fetcher-fast']","time":"datetime.datetime(2024, 12, 2, 19, 45, 19)"},
{"count":"1L","exchange":"''","original-expiration":"'3600000'","queue":"'fetcher-delay'","reason":"'expired'","routing-keys":"['fetcher-delay']","time":"datetime.datetime(2024, 12, 2, 19, 43, 16)"}
],
x-first-death-exchange: "",
x-first-death-queue: "fetcher-delay",
x-first-death-reason: "expired",
x-mc-name: "_on_input_message",
x-mc-retries: 2,
x-mc-what: "Retry('skipped due to recent connection failure')",
x-mc-when: "1733164994.9960396",
x-mc-where: "/app/indexer/workers/fetcher/tqfetcher.py:461",
x-mc-who: "fetcher"
}
}

**Comment by philbudne** [comment]
Yeah, I've hit my head on this when trying to use the qutil
"dump_archives" command on the fetcher-fast queue, or on any -delay
queue.

Maybe the right place to make datetimes JSON-safe (by calling
.isoformat()) might be between these two lines in the `dump_archives`
method in qutil.py:

```
extras = {"rabbitmq_headers": properties.headers}
aw.write_story(story, extra_metadata=extras, raise_errors=False)
```

Another possibility is to implement what the comment
just above those two lines suggests?

```
# XXX just save headers that start with "x-mc-"?
```

Since the data in "rabbitmq_headers" data isn't used for anything
(but it _COULD_ be useful in autopsy/forensic examination)

PR #361, opened 2024-12-22
qutil improvements

philbudne:
fix https://github.com/mediacloud/story-indexer/issues/355 by only dumping x-mc- headers (as noted in comments)

Add --really-queue and --ignore-domain options

### DIFF
--- a/indexer/scripts/qutil.py
+++ b/indexer/scripts/qutil.py
@@ -14,6 +14,7 @@ import sys
 from logging import getLogger
 from types import FrameType
 from typing import Callable, List, Optional, cast
+from urllib.parse import urlparse
 
 # PyPI
 from pika import BasicProperties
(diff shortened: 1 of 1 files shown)


pgulley (approved):
could you add a note on what the alternative to 'really queue' is? why would we want not to really queue?

philbudne:
@pgulley without the option it's a "dry run" (tests whether the file is valid).  I wanted to be able to test whether the file was valid before committing to actually queuing it (and encourage that behavior).  TOTALLY open to better naming!!!

pgulley:
Calling the flag '--dry-run,' or  '--test-run,' and defaulting it to true feels less convoluted to me, as long as there's a comment in the help message explaining that flipping it to false is necessary to actually queue

philbudne:
A humorous contranym for --dry-run would be --wet-run!

philbudne:
replaced by https://github.com/mediacloud/story-indexer/pull/412

Issue #353, opened 2024-11-24
Re-filling the feb-may 2022 "dip" using canonical URL extraction


philbudne:
Given the successful recovery using "blind fetching" S3 objects and using the extracted canonical URL, I suggested we might want to do the same thing for the period in 2022 (approx 2022-01-25 thru 2022-05-05?) where Xavier fetched all the S3 objects, looking only at the RSS files to extract URLs that we tried to fetch again (and found significant "link rot").

It looks like the researchers would prefer filling out 2022 to working on prior years (2019 and earlier).

To see if it might be a savings (assuming my memory that access to S3 is free from EC2 instances is correct) to scan the S3 objects from an EC2 instance, and pack them (without ANY further processing, including checking if the file has a canonical URL) into WARC files,  I threw together a packer.py script with bits cribbed from hist-fetcher and parser (for RSS detection).

To get 100 HTML files it scanned 767 S3 objects (some non-existent, some 36 bytes or smaller which almost certainly contain a message that says it was a duplicate feed download), downloading a total of 5676225 bytes (avg 7400 bytes/obj scanned), and writing a WARC file that's 2726394 bytes (48% the size), so it might be worthwhile (with more consideration and math).

Then I created a t4g.nano instance ($0.0042/hr) to see how much faster downloads are from inside AWS, and it took about half the time (23 seconds vs 48seconds from ifill).  That doesn't include additional time for the EC2 instance to copy the WARC file to S3.

Further data points:

My initial estimate (working only from the previous run of a one month period, and factoring in halving the speed to avoid hogging UMass bandwidth) was that it could take 56 days.

Poking around in the S3 bucket, it looks like the object ID range is about 113 million objects to be scanned; at 50 downloads/second (current historical ingest rate with 6 fetchers), that looks like it could be only 26 days.

So six "packers" (each given a share of the object ID range) running in EC2 at 33 obj/second is 200 obj/second, and 113Mobj divided by 200 obj/sec looks to be about a week of EC2 time.

A t3a.xlarge instance (4 AMD CPUs) is $0.15/hr, which would be $25 for a week (not counting EBS costs for the root disk).

Amazon pricing usually doubles for a doubling in resources, so the total price might be the same for different instance sizes, the instance size just determines the speed (assuming there isn't some other bottleneck).

With the 7400 bytes/obj number from above, at 113M objects, that's 836GB of download to transfer the raw objects, 
The WARC file came in at 3555 bytes/object or 402GB to download.

[...]

PR #364, opened 2025-01-27
Add ability to remove archive documents from ElasticSearch using original URL 

m453h:
This PR modifies the [Arch Eraser](https://github.com/philbudne/story-indexer/blob/arch-eraser-sketch/indexer/scripts/arch-eraser.py) sketch to allow the deletion of archived documents from ElasticSearch.

The python script can be invoked as

```
 python arch-eraser.py b2://mediacloud-indexer-archive/2024/12/14/mchist2022-20241214235742-291-9d4ffe800d4a.warc.gz --rabbitmq-url="-" --elasticsearch-hosts="http://localhost:9200" --indices="mc_search-000001"
```

Since the documents are stored with a certain  prefix (hist2022) within a specified date as discussed [here](https://github.com/mediacloud/story-indexer/issues/353#issuecomment-2593349155) the PR includes a bash script that generates a pattern that can be used to match documents with a certain remote URL pattern to ease how we invoke the arch-eraser.  This script can be executed using the following format:
```
./bin/run-arch-eraser.sh <START_DATE> <END_DATE> <REMOTE_URL> '<OTHER OPTIONS>'
```
Example:
```
./bin/run-arch-eraser.sh 2024/12/14 2024/12/16 "b2://mediacloud-indexer-archive/{pattern}/mchist2022" '--rabbitmq-url="-"  --elasticsearch-hosts=http://localhost:9200  --indices=mc_search-000001' 
```

Will append: `b2://mediacloud-indexer-archive/2024/12/14/mchist2022 b2://mediacloud-indexer-archive/2024/12/15/mchist2022 b2://mediacloud-indexer-archive/2024/12/16/mchist2022` as args when we invoke the arch-eraser script.

The script supports both bulk and individual record removal from Elasticsearch. I tested it on my local Elasticsearch instance with approximately 500,000 documents, removing URLs extracted from 50 WARC files containing around 250,000 records. Over 10 iterations of the same task, the following average time was observed:
- Bulk deletion: 11.55 seconds
- Individual records deletion : 271.09 seconds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ docker/deploy.log
 docker/docker-compose.yml
 docker/docker-compose.yml.dump
 docker/docker-compose.yml.save-*
+*.sqlite3
--- a/bin/run-arch-eraser.sh
+++ b/bin/run-arch-eraser.sh
@@ -24,7 +24,7 @@ increment_date() {
     esac
 }
 
-if [ $# -ne 3 ]; then
+if [ $# -lt 3 ]; then
     echo "Usage: $0 [start_date] [end_date] [pattern]" >&2
     exit 1
 fi
@@ -32,6 +32,8 @@ fi
 start_date="$1"
 end_date="$2"
 pattern="$3"
+shift 3
+other_params="$*"
 
 if ! is_valid_date "$start_date" || ! is_valid_date "$end_date"; then
     echo "Error: Invalid date format. Use YYYY/MM/DD" >&2
@@ -50,7 +52,7 @@ convert_date_to_int() {
     fi
 }
 
-output_string="--rabbitmq-url='-' "
+output_string=""
 start_date_int=$(convert_date_to_int "$start_date")
 end_date_int=$(convert_date_to_int "$end_date")
 
@@ -61,4 +63,4 @@ while [ "$start_date_int" -le "$end_date_int" ]; do
     start_date_int=$(convert_date_to_int "$start_date")
 done
 
-run_python indexer.scripts.arch-eraser $output_string
+run_python indexer.scripts.arch-eraser ${output_string} ${other_params}
--- a/bin/run-arch-eraser.sh
+++ b/bin/run-arch-eraser.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+
+. bin/func.sh
+
+is_valid_date() {
+    case "$1" in
+        [0-9][0-9][0-9][0-9]/[0-9][0-9]/[0-9][0-9]) return 0 ;;
+        *) return 1 ;;
+    esac
+}
+increment_date() {
+    date_cmd=$(command -v date)
+    case "$(uname)" in
+        Darwin)
+            "$date_cmd" -j -v+1d -f "%Y/%m/%d" "$1" +"%Y/%m/%d"
+            ;;
+        Linux)
+            "$date_cmd" -d "$1 + 1 day" +"%Y/%m/%d" 2>/dev/null
+            ;;
+        *)
+            echo "Unsupported Environment" >&2
+            return 1
+            ;;
+    esac
+}
+
+if [ $# -ne 3 ]; then
+    echo "Usage: $0 [start_date] [end_date] [pattern]" >&2
+    exit 1
+fi
+
+start_date="$1"
+end_date="$2"
+pattern="$3"
+
+if ! is_valid_date "$start_date" || ! is_valid_date "$end_date"; then
+    echo "Error: Invalid date format. Use YYYY/MM/DD" >&2
+    exit 1
+fi
+
+convert_date_to_int() {
+    input_date="$1"
+    if [ "$(uname)" = "Darwin" ]; then
+        date -j -f "%Y/%m/%d" "$input_date" +"%Y%m%d" 2>/dev/null
+    elif [ "$(uname)" = "Linux" ]; then
+        date -d "$input_date" +"%Y%m%d" 2>/dev/null
+    else
+        echo "Unsupported OS" >&2
+        return 1
+    fi
+}
+
+output_string="--rabbitmq-url='-' "
+start_date_int=$(convert_date_to_int "$start_date")
+end_date_int=$(convert_date_to_int "$end_date")
+
+while [ "$start_date_int" -le "$end_date_int" ]; do
+    current_url=$(echo "$pattern" | sed "s|{pattern}|$start_date|g")
+    output_string="${output_string}${current_url} "
+    start_date=$(increment_date "$start_date")
+    start_date_int=$(convert_date_to_int "$start_date")
+done
+
+run_python indexer.scripts.arch-eraser $output_string
--- a/indexer/queuer.py
+++ b/indexer/queuer.py
@@ -212,6 +212,17 @@ class Queuer(ShufflingStoryProducer):
         args = self.args
         assert args
 
+        if fname[0] == "@":
+            # implement "indirect file" (file containing file names)
+            # NOTE! paths read from indirect files are NOT interpreted
+            # as relative to the path of the indirect file.
+            logger.info("indirect file %s", fname)
+
+            f = self.open_file(fname[1:])
+            for line in f:
+                self.maybe_process_file(line.decode().rstrip())
+            return
+
         if args.test:
             logger.info("maybe_process_file %s", fname)
             return
--- a/indexer/scripts/arch-eraser.py
+++ b/indexer/scripts/arch-eraser.py
@@ -25,48 +25,39 @@ with some available from B2 as well (may be cheaper to fetch from B2).
 
 2. From RSS files blindly extracted from S3 (ignoring the HTML files!) into CSV files of URLs
 
-================ index mc_search-00002 created 2024-03-24T01:07:19.076Z
+Stories in index mc_search-00002 and mc_search-00003
 
 All files on S3, some on B2 (starting 2024/05/31)
 
 arch prefix     start           end             archives
-mccsv           2024/05/22 -> 2024/06/22        S3/(B2)
-mc[rss]         2024/05/27 -> 2024/06/22        S3/(B2) [1]
+mccsv           2024/05/22 -> 2024/06/27        S3/(B2)
+mc(rss)         2024/05/27 -> 2024/06/20        S3/(B2) [1]
+mcrss           2024/06/20 -> 2024/08/16        S3/B2
 
 (B2) means some of the date range on B2
 
-[1] files named mcrss- start 2024/06/20.
-    files written from 2024/05/27 thru 2024/06/20 start with mc-
-    (see below):
+[1] initial WARC files from RSS files written from 2024/05/27 thru
+    2024/06/20 start with mc- (see below):
 
     THESE SHOULD BE VERIFIED!!! The "via" field in the metadata
     should indicate how the URL was obtained!
 
     dates                   container name (in WARC filename)
-    2024/05/27-2024/05/28   cf94b52abe5a
-    2024/05/29-2024/06/04   cefd3fdce464
-    2024/06/05              0c501ed61cf4
-    2024/06/05              446d55936e82
+    2024/05/27-2024/05/28   cf94b52abe5a        S3 [154 files]
+    2024/05/29-2024/06/04   cefd3fdce464        S3 [882 files]
+    2024/06/05              0c501ed61cf4        S3 & B2 [497 files]
+    2024/06/05              446d55936e82        S3 & B2 [27 files]
     2024/06/05              cefd3fdce464
     2024/06/06-2024/06/09   0c501ed61cf4
-    2024/06/09              7e1b47c305f1
+    2024/06/09              7e1b47c305f1        S3 & B2 [1 file]
     2024/06/11-2024/06/20   6c55aaf9daaa
-
-================ mc_search-000003 created 2024-06-22T01:09:36.337Z
-
-arch prefix     start           end             archives
-mccsv           2024/06/22 -> 2024/06/27        S3/(B2)
-mcrss           2024/06/22 -> 2024/08/16        S3/(B2)
-
-(B2) means some of the date range on B2
-(S3) means some of the date range on S3
 
 ================
 
 This is based on the "Queuer" class, which reads both local and remote
 input files, and keeps track of which files have been processed.
 
-No queues are involved, so this code should be runnable outside of Docker.
+No queues are involved (provide any value for --rabbitmq-url or RABBITMQ_URL)
 
 The "tracker" uses SQLite3 (**), and should be multi-process safe,
 although this application may experience more contention (SQLite3 does
--- a/indexer/scripts/arch-eraser.py
+++ b/indexer/scripts/arch-eraser.py
@@ -75,21 +75,33 @@ file (or files), and use the "indirect file" feature in the queuer
 to read the files of URLs.
 """
 
+import argparse
 import logging
-from typing import BinaryIO
+import time
+from typing import BinaryIO, List, Optional
 
+from elasticsearch import Elasticsearch
+
+from indexer.elastic import ElasticMixin
 from indexer.queuer import Queuer
 from indexer.story_archive_writer import StoryArchiveReader
 
 logger = logging.getLogger("arch-eraser")
 
 
-class ArchEraser(Queuer):
+class ArchEraser(ElasticMixin, Queuer):
     APP_BLOBSTORE = "HIST"  # first choice for blobstore conf vars
     HANDLE_GZIP = False  # StoryArchiveReader handles compression
 
     # don't want to talk to RabbitMQ, but too much work
     # to refactor Queuer into a FileProcessor add-in
+
+    def __init__(self, process_name: str, descr: str):
+        super().__init__(process_name, descr)
+        self.is_batch_delete: bool = False
+        self.keep_alive: str = ""
+        self.fetch_batch_size: Optional[int] = None
+        self.indices: str = ""
 
     def qconnect(self) -> None:
         return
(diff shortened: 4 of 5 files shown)


thepsalmist (commented):
Mostly LGTM 👍

indexer/scripts/arch-eraser.py
```diff
@@ -0,0 +1,227 @@
...
+                if bulk_actions:
+                    es.bulk(index=self.indices, body=bulk_actions)
+                    total_deleted += len(bulk_actions)
+                search_after = hits[-1]["sort"]
+        except Exception as e:
+            logger.exception(e)
+        finally:
+            if total_deleted != len(urls):
+                logger.warning(
+                    f"Mismatch in document deletion count: [{total_deleted}] deleted out of [{len(urls)}] expected."
```
thepsalmist, line 197: In some places we're logging using the `%` style, while some others we using `f-strings`, can standardise to `%`

indexer/scripts/arch-eraser.py
```diff
@@ -0,0 +1,227 @@
...
+        self.keep_alive = self.args.keep_alive
+        self.is_batch_delete = self.args.is_batch_delete
+
+    def delete_documents(self, urls: List[Optional[str]]) -> None:
+        es = self.elasticsearch_client()
+        pit_id = None
+        total_deleted = 0
+        try:
+            pit_id = es.open_point_in_time(
+                index=self.indices, keep_alive=self.keep_alive
```
thepsalmist, line 157: I see, [open_point_in_time](https://elasticsearch-py.readthedocs.io/en/stable/api/elasticsearch.html#elasticsearch.client.Elasticsearch.open_point_in_time) allows `_all` or an empty string of indices. Does this presume we know the exact indices to delete from?

philbudne: My notes indicate that most of the stories are in mc_search-00002 but that some are in mc_search-00003

philbudne (changes requested):
I think it would be safer to have the script prepare files with lists of WARC files (to be used with `@file` on the command line); rather than running the delete: that way the lists can be prepared and inspected before the deletion is done!

indexer/scripts/arch-eraser.py
thepsalmist: Cool, if there's certainty in all being in 0002 and 0003

philbudne (approved):

indexer/scripts/arch-eraser.py
m453h: Yes and we can also use '_all' as a parameter when we call the script in a different scenario

m453h:
I have run the script to get the WARC files from B2 as follows:
```
./bin/run-arch-eraser.sh --generate-erase-list 2024/12/14 2024/12/31 "b2://mediacloud-indexer-archive/{pattern}/mchist2022" --output-file=arch-eraser/erase-list.txt 
```
 Attached is the list of files we expect to process and delete from Elasticsearch:

[erase-list.txt](https://github.com/user-attachments/files/18592035/erase-list.txt)

philbudne:
@m453h the list should include more files (ones with just mc- prefix and ones with mcrss- prefix), but I think that discussion can take place in issue https://github.com/mediacloud/story-indexer/issues/353

I think we should have many eyes examine the generated lists before doing the erasure: @pgulley should be add some assignees to issue 353 (above) as we continue the process?

indexer/queuer.py
```diff
@@ -189,8 +191,19 @@ def maybe_process_files(self, fname: str) -> None:
...
-                    for key in sorted(bs.list_objects(prefix), reverse=True):
-                        self.maybe_process_file(f"{scheme}://{bs.bucket}/{key}")
+                    assert self.args
+                    output_file: Optional[str] = self.args.output_file
+                    context: ContextManager[Optional[object]] = nullcontext()
+                    with (
+                        open(output_file, "a")
+                        if output_file
+                        else context
+                    ) as file:
```
kilemensi, line 201: Can't the `if` be checked outside `with`? Haven't worked with `ContextManager` but doesn't something like 👇🏽 work?
```python
    if self.args.output_file:
        output_file = open(self.args.output_file, "a")
    else:
        output_file = nullcontext()

    with output_file as file:
```

Multiline with is a bit hard to follow.

indexer/queuer.py
```diff
@@ -189,8 +191,19 @@ def maybe_process_files(self, fname: str) -> None:
...
+                        for key in sorted(bs.list_objects(prefix), reverse=True):
+                            if output_file:
```
kilemensi, line 203: Why are we checking `if output_file` here when the outer `with` should have done that?

bin/run-arch-eraser.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/sh
```
kilemensi, line 1: I take it I'm the only one who prefers python to bash scripts 😢 ... I guess I'll go cry in a corner.

kilemensi (approved):
🚀

indexer/queuer.py
m453h: Sure I made a few minor typing adjustments to get MyPy to play nice

thepsalmist (approved):
🚀 LGTM

bin/run-arch-eraser.sh
```diff
@@ -0,0 +1,151 @@
...
+    echo "  --erase                         Deletes records from Elasticsearch."
+    echo "  -h, --help                      Show this help message."
+    echo "  -h --generate-erase-list        Show help for the --generate-erase-list action."
+    echo "  -h --erase                      Show help for the --erase action."
+    exit 0
+}
+
+# Handle help flag
+if [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+    if [ "$2" = "--generate-erase-list" ]; then
```
philbudne, line 95: Requiring an ordered pair of (positional?) options is unusual to say the least.

I've suggested (at least once) that generating the lists and actually erasing be separate operations to avoid nasty surprises due to finger (or brain slips).  Simpler is better.

.gitignore
```diff
@@ -14,3 +14,5 @@ docker/deploy.log
 docker/docker-compose.yml
 docker/docker-compose.yml.dump
 docker/docker-compose.yml.save-*
+*.sqlite3
+arch-eraser/*
```
philbudne, line 18: Are these needed?

The "tracker" should be writing sqlite db files inside the data directory, whose contents is already ignored.

Is there a top-level arch-eraser directory?  It should be inside the data directory (see above).

indexer/queuer.py
```diff
@@ -189,8 +192,18 @@ def maybe_process_files(self, fname: str) -> None:
                     # more likely a queuer might actually need
                     # multiple keys (eg; reading CSVs from one bucket
                     # that reference objects in another bucket).
-                    for key in sorted(bs.list_objects(prefix), reverse=True):
-                        self.maybe_process_file(f"{scheme}://{bs.bucket}/{key}")
+
+                    if self.args and self.args.output_file:
```
philbudne, line 196: Why is the output file option needed?  `maybe_process_file` logs file names before `process_file` is called *AND* records progress in the "tracker" (we usually just query the sqlite3 file for long-running jobs to watch progress).

philbudne (changes requested):
Some suggestions on simplifications

philbudne:
As an old Unix programmer I think programs should do one thing, and as
an old Python programmer, I think there should be one way to do things.

My concerns are:

I'd like to be able to review what is going to be removed: to see the
files that contain the names of the files that will be processed, not
the list of files that will be processed if almost exactly the same
command is later entered.

It's incredibly easy to mistype a command.  I recently accidentally
shut down a server on a friday night by typing the wrong thing in the
wrong window.

Having there be two ways to do it (using the reviewed file, or
re-entering the command) reduces my confidence that the reviewed lists
are going to be the ones actually used.

I'll leave the final call to @pgulley as the client.

To summarize: When serious/important/delicate operations are done,
I think it's best to have them done in repeatable, self documenting
ways.

For example, we're about to set up eight servers.  I think it's
preferable to automate the process (even if it takes longer) so that
it's done consistently, so if (or when) it needs to be redone, the
same result is reached.

m453h:
> As an old Unix programmer I think programs should do one thing, and as an old Python programmer, I think there should be one way to do things. My concerns are: I'd like to be able to review what is going to be removed: to see the files that contain the names of the files that will be processed, not the list of files that will be processed if almost exactly the same command is later entered. It's incredibly easy to mistype a command. I recently accidentally shut down a server on a friday night by typing the wrong thing in the wrong window. Having there be two ways to do it (using the reviewed file, or re-entering the command) reduces my confidence that the reviewed lists are going to be the ones actually used. I'll leave the final call to @pgulley as the client. To summarize: When serious/important/delicate operations are done, I think it's best to have them done in repeatable, self documenting ways. For example, we're about to set up eight servers. I think it's preferable to automate the process (even if it takes longer) so that it's done consistently, so if (or when) it needs to be redone, the same result is reached.

Hey @philbudne I agree with the concerns raised and believe its important to improve on the current approach. Before moving forward with these improvements, I want to ensure I’m not overlooking anything. I’m proposing the following changes to address the concerns:

1. Use the `SQLite3FileTracker` to store the names of the files to be deleted this would mean I need to either modify the SQLite schema to also store the full path of the file to be deleted or store the full path instead of just the name. Any audit of the records that will be deleted should be done on the SQLite database instead of using an external txt files.

 2. Implement separate processes that will be invoked by two different bash script. 
     - The first process will populate the database with the list of the files to be deleted (Files will have a NOT_STARTED status)
     - The second process will be responsible for the actual deletion, it will read from the database and call the process_file method with the delete option. This will ensure that only WARC files in the list of files to be deleted stored in SQLite will be used for deletion from ElasticSearch. Additionally, for extra safety I could perhaps include another status that would imply that we have verified the files in  SQLite, ensuring that the second process only starts after confirmation.

Please let me know if this approach makes sense and effectively addresses the key concerns with the current implementation.

pgulley:
@m453h Thanks for laying out that approach- I think using sqlite in the middle sounds fine, and having a singular spot in between the two processes helps make sure that things are done in order. @philbudne , does this address your concerns? I'd love to move on this task asap so we can clear our plates for the new servers.

philbudne:
I had mentioned querying SQLite3 based Tracker files (used in the
Queuer framework to allow multiple active processes to atomically
claim ownership of one of many input files (WARC, RSS, CSV that might
be local or on S3 or B2) as making added code to write an output file
containing a list of files that had already been processsed.

I wasn't suggesting using SQLite to track the individual URLs to be
removed.  SQLite doesn't do row locking, only full table locking, so
I'm concerned that it's most suitable for tracking files that contain
batches of work, not individual stories...

I think we're now talking about a five step process:

Step 1 can be Michael's script, preferable outputting a file
   for the three ways we filled in the "dip":
      a. From CSVs generated by reading RSS files from S3
	  (WARC files starting with mccsv-)
      b. From RSS files, generating WARCs starting with mc-
	  (these are the ones that need the closest examination!)
      c. From RSS files, generating WARCs starting with mcrss-

Step 2 "archive-eraser" renamed archive-lister(?);  As many copies as
   desired can be run, either with @file-a @file-b @file-c (and they
   will compete for input files using the SQLite3 based tracker) or by
   running three groups of one or more with each of the input files.
   The "process_story" method would write the URL to a different file
   specified with -o for each invocation.  Having one output file
   for each WARC would give additional information, and help split
   up the work for the next step.

Step 3 the eraser: Could be based on the csv-queuer where the
  "process_story" method does the erasing, and using the Tracker
  mechanism to compete over "url list" files) so that the number
  of workers at any time can be varied.

Step 4: Getting a list of "mchist2022-" warc generated in December into a file
   or files using Michael's script.

Step 5: Copying the files from 4 into the a subdirectory in the
  arch-indexer stack "/app/data" (ie named mchist2022) and starting
  a stack with '-T archive' and '-I data/mchist2022'
  (fuzzy on this step, I haven't run an archive stack in 
  almost EXACTLY a year!!)

m453h:
Thanks @philbudne for the description of the steps, I have made the changes to the implementation we now have 3 separate scripts (I did rename `arch-eraser.py `to `arch-lister.py` which makes more sense since the class only does listing of files):
Here is a brief description of the scripts:
- `run-arch-warc-lister.sh`: this receives start date, end date and search pattern as input and outputs a file that contains a list of WARC files to be processed, it will be used for Step 1 and Step 4
- `run-arch-url-lister.sh`: this receives an indirect file that contains a list of WARC files to process (generated by `run-arch-warc-lister.sh`) and outputs a list of URLs to be deleted (it produces one txt file per one WARC file it has processed by default)
- `run-arch-eraser`: this receives a path containing files with list of URLs to delete  (generated by `run-arch-url-lister.sh`) from elasticsearch and does the actual deletion.

I hope this significantly improves the deletion process, I have also added some few safety features like retries when deletion from Elasticsearch fails.

kilemensi (commented):
👍🏽 

---

Since we've made some major changes, can you now include how the various outputs look like (may be by updating the PR description)

indexer/workers/fetcher/csv_queuer.py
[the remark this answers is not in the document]
m453h: I'm importing it in the [arch-eraser.py](https://github.com/mediacloud/story-indexer/blob/16e3e77fb5e9115be40cfb745781c4c240e21c61/indexer/scripts/arch-eraser.py#L11C1-L12C1)... I had to remove the hyphen to have a consistent way we are importing modules i.e. I was avoiding using [importlib](https://docs.python.org/3/library/importlib.html)

indexer/scripts/arch-eraser.py
```diff
@@ -0,0 +1,257 @@
...
+            logger.exception("Error processing document %s: %s", hit.get("id"), e)
+
+    def _bulk_delete(self) -> None:
+        if not self.buffer:
+            return
+        sec = 1 / 16
+        while True:
+            try:
+                assert self.es_client
+                success, _ = bulk(
```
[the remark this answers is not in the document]
m453h: 1. This could possibly work and we could completely rely on the retry behavior that we get from the helper function.
However, one issue I encountered was that when a request times out, the client doesn’t retry the request and instead throws an exception. I did a bit more reading on this and one way would be to set `retry_on_timeout=True` by overriding the method we are using to create an instance of the Elasticsearch client. I'll make some tests to be sure we'll get consistent behaviour. 

2. I'm not exactly sure if deleting one document using the bulk helper would have the same impact as deleting one document using the `delete` method, I had kept it as just an alternative if it happens that the bulk operations would be failing more often when we do the actual deletion (one idea we had initially discussed was to first try to do a bulk delete operation and if it fails to maybe try to try calling a delete operation per each individual document in our delete batch), but we could completely remove this path for the sake of simplicity.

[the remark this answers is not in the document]
kilemensi: ... and to be 100% clear, I'm not saying the bulk update is the better option (other than it appears to have retry logic built-in); I'm saying it's best we evaluate and pick one method that will work well for our case(s) and optimise for that rather than having 2 methods that may behave completely different at runtime.

indexer/workers/fetcher/csv_queuer.py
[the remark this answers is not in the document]
kilemensi: Should we then have an issue to track file naming that may lead to import errors?

indexer/scripts/arch-eraser.py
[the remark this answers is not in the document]
m453h: Sure, from the initial tests I did locally, I would advise we go for the bulk delete option (It significantly reduces the time required to delete documents)

indexer/scripts/arch-eraser.py
```diff
@@ -0,0 +1,257 @@
...
+from elasticsearch import Elasticsearch
+from elasticsearch.helpers import bulk
+
+from indexer.elastic import ElasticMixin
+from indexer.workers.fetcher.csv_queuer import CSVQueuer
+
+logger = logging.getLogger("arch-eraser")
+
+
+class ArchEraser(ElasticMixin, CSVQueuer):
```
philbudne, line 16: When I mentioned csv-queuer.py, I meant to grab/copy it, as an example of a very simple Queuer, one that reads a text file containing URLs, not that it should subclass it!

In general, my preference is to name scripts (things that will be run rather than imported) with hyphens in their names, and if some functionality needs to be shared, then it belongs in the top "indexer" directory, and should have underscores in its name for importing.

indexer/scripts/arch-eraser.py
```diff
@@ -0,0 +1,257 @@
...
+            )
+        except Exception as e:
+            logger.exception(e)
+        finally:
+            self._close_pit()
+
+    def _fetch_documents_to_delete(self, urls: List[str]):
+        try:
+            query = {
+                "size": self.fetch_batch_size,
```
philbudne, line 142: I'm surprised to see a query!

If it's just to find the ID of the document to delete, then you can use `mcmetadata.urls.unique_url_hash()` on the URL.

Any reason to use the `original_url` field (below)?

It ended up in ES because it's one of the outputs of mcmetadata, and in normal operation it should be identical to `url`, and I don't think it's actually used by anything (if it was possible to remove fields in the move to the new cluster, it would be one of two I'd suggest tossing, the other is full_language)!

indexer/workers/fetcher/csv_queuer.py
philbudne, line 1: See above comments.  Please leave the name alone, and just copy the file!

philbudne (changes requested):

indexer/workers/fetcher/csv_queuer.py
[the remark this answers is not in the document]
m453h: This is no longer necessary as described [here](https://github.com/mediacloud/story-indexer/pull/364#discussion_r1957039043).

indexer/scripts/arch-eraser.py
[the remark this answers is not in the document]
m453h: @kilemensi one limitation I have faced while using the combination of `retry_on_status`, `max_retries`, `max_backoff` is that the deletion process will halt with an exception when there is complete network failure and Elasticsearch fails to return a response. I believe we still need to retry in this scenario after waiting for a few seconds before exiting.

indexer/scripts/arch-eraser.py
m453h: Thanks, for the added context ! @philbudne  I have made the change to use the `ID` constructed from `mcmetadata.urls.unique_url_hash()`

Regarding why we have a query, it is to retrieve both the `index` and `ID` of the document, as Elasticsearch requires this information for deletion. Since the stories could be in  either `mc_search-00002` or `mc_search-00003`, I thought retrieving the story by ~~original_url~~ `ID` to identify its index before constructing the list of stories for bulk_delete would be the most appropriate approach.

An alternative to prefetching the stories to determine their index would be to use the [delete_by_query](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html) API. However, this can potentially use the [scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html) when handling large datasets, adding more load to Elasticsearch. That's why I opted for a query using search_after and PIT. 

I’m definitely open to exploring any other approach you think could improve this!

philbudne: @m453h  haven't studied delete_by_query, but only one document should EVER match the "id" within a single index, and we try not to index a URL more than once by doing an id query against all indices before trying to import a story, so if the delete_by_query is limited to index ...02 and ...03 the worst case is two documents.

My concern is that the separate lookup and delete will have longer latency (two round trips thru the API) and more impact on the ES servers (locating the document twice).  On the other hand I could be worrying about nothing!

philbudne (approved):
Looks good!

m453h:
Hey @pgulley could you have a final look to confirm  if this is ready for merging ? I believe after merging the next steps would be to confirm if the list of WARC files I generated on #353 is correct and decide when it would be appropriate to run the script for the actual deletion

pgulley (approved):
yes, looks great to me!

Issue #366, opened 2025-02-19
Investigate elasticsearch configuration using ansible


philbudne:
Keeping notes about what I learn here...
The issue about cluster configuration is https://github.com/mediacloud/story-indexer/issues/365

philbudne:
The first thing I found is an archived github repo: https://github.com/elastic/ansible-elasticsearch?tab=readme-ov-file
which doesn't support ES 8.x: https://github.com/elastic/ansible-elasticsearch/blob/main/docs/8x-support.md#what-is-not-working

The second thing I found is stuff to manage elastic (snapshots, reindex) using ansible, so not applicable in THIS situation:
https://galaxy.ansible.com/ui/repo/published/community/elastic/docs/
Tho it's possible https://galaxy.ansible.com/ui/repo/published/community/elastic/content/module/elastic_cluster_settings/
might do what we want/need.

The third thing: https://github.com/geerlingguy/ansible-role-elasticsearch
"An Ansible Role that installs Elasticsearch on RedHat/CentOS or Debian/Ubuntu."
Author Information:  This role was created in 2014 by Jeff Geerling Ansible for DevOps

philbudne:
Looking at forks of archived elastic role (the first link in this issue):
https://github.com/elastic/ansible-elasticsearch/forks
there are 857 forks(!!), 16 modified in the last year:
https://github.com/elastic/ansible-elasticsearch/forks?include=active&page=1&period=1y&sort_by=stargazer_counts

Searching for elasticsearch 8 ansible:

[ElasticSearch 8.x installation on Ubuntu 20.04 LTS with Ansible](https://www.youtube.com/watch?v=X0BHxZIZMNw)

https://medium.com/@abdullah.pelit/installation-of-elk-stack-version-8-9-with-ansible-aaec8a0f2765

philbudne:
Of the forks of the original elastic ansible role (first link), these seem to have been updated for 8.x:
https://github.com/thuync/ansible-elasticsearch
https://github.com/ES-MAD-GEMMA-ROADMAP/ansible-elasticsearch

Unclear how many (if any) changes are ACTUALLY needed for v8!

Many have changes for changes to ansible "include" directive.

pgulley:
Great work compiling resources! So it seems we're stuck using a fork of ansible to get this done- I say just pick one that looks good, I'm not sure what input I have that might help differentiate, other than it should probably be minimally different from ansible core so we can keep up with changes there.

[...]

PR #369, opened 2025-03-10
Elasticsearch Ansible Setup

thepsalmist:
This PR introduces Ansible playbooks and roles for deploying ad managing an Elasticsearch cluster.
This builds off from the discussion on #366 and the draft at #367


conf/ansible/inventories/group_vars/vault.yml
DavidTheProgrammer, line 1: This directory ideally should only contain variables shared by specific groups and since we don't have a group named "vault", I'd recommend we move this one level up to the `inventory` level. The rationale is that this is the vault used for this particular inventory.

conf/ansible/inventories/hosts.yml
DavidTheProgrammer, line 1: I have found the alternative directory layout located in the Ansible best practices that seperates the inventories based on environment to work quite well in the past. Reduces filters needed and allows us to actually use `all` group for things like agent installation that needs to happen on all hosts while only affecting specific environments.

https://docs.ansible.com/ansible/2.8/user_guide/playbooks_best_practices.html#alternative-directory-layout

conf/ansible/playbooks/es-install.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Install and configure Elasticsearch cluster
+  hosts: elasticsearch
+  vars_files:
+    - ../inventories/group_vars/vault.yml
+    - ../roles/elasticsearch/vars/{{ env | default('production') }}.yml
```
DavidTheProgrammer, line 6: Can we default to `staging`? I think that's a safer default.

conf/ansible/playbooks/es-install.yml
```diff
@@ -0,0 +1,38 @@
...
+  become: true
+
+  roles:
```
DavidTheProgrammer, line 9: > Use include_role instead of the traditional roles: section

Following this article we should prefer `include_roles` instead of the `roles` section. I think it's okay.

https://www.ansiblejunky.com/blog/ansible-101-standards/#playbooks

conf/ansible/playbooks/es-uninstall.yml
DavidTheProgrammer, line 1: Can we move these to a standalone role and include it in this playbook instead of having the tasks written directly? This also comes from the recommendations in the Ansible 101 article.

https://www.ansiblejunky.com/blog/ansible-101-standards/#playbooks

DavidTheProgrammer (commented):
Looks good, I've left a few comments on the general structure and I have a couple of questions about the role we're using, is it something we've hand-rolled or it's available on Ansible Galaxy? It seems to be environment dependent with the vars files and that is setting off some flags for me as roles should not be environment dependent. Is there any reason we can't use a 3rd party installation modules for ES 8+ Like this one: https://galaxy.ansible.com/ui/standalone/roles/geerlingguy/elasticsearch/install/ ?

philbudne:
I started the ansible files.

Regarding selection of a role to install ES: I surveyed the available
roles, and I'm pretty sure the geerlingguy module was one I glanced
at.  I didn't keep notes on why I rejected it (the fact that the docs
don't show ES 8.x support, and it requires a separate module for Java
install probably didn't weigh in it's favor).

I ended up choosing between two forks of the Elastic role that had
been adapted for ES8, selecting the more lightly modified one, and
forked it to make one or two changes necessary to get it to run,
keeping it as a separate (mediacloud) repo because (a) it had it's own
licence, and (b) I didn't want to get too deep into modifiying it.

Some of my thoughts / principles:

To keep the work compact, and avoid spreading stuff out into too many
files: Keeping the node names ONLY in the YAML inventory file, with
per-group settings in the same file, allowing testing by selecting an
alternate inventory group.

I regard ansible as a necessary evil (hateful in many ways, tho the
best/only game in town).  Past experience has shown me that data
organization is key (being a data driven solution), and that it's easy
to make a twisted mess that's hard to untangle (and noone wants to
touch because it's so easy to break).

Like in many things, I'm not a purist or true believer.  My experience
is that it's likely that any ansible file written a few years ago
won't run as-it today, and I have no reason to believe the same won't
be true of today's files in five years: I tried to avoid things that
caused overt warnings, but I view "best practices" and "future
proofing" as at best unnecessary, or at worst harmful to clarity and
simplicity (making something that gets the job done, and future people
can understand and adapt quickly) as opposed to maximally reusable by
others.

Trying to be a "hero" was an ABSOLUTE non-goal.  I quickly realized
that my install of statsd-agent.py SHOULD have been written as a
sequence of tasks, rather than a slab of shell commands, AND that if I
made it a role, it could be reused to install other tools in the same
repo that could be used to rebuild other servers (ie; tarbell, the web
server), *BUT* I didn't do either of those things; it simply isn't
worth the investment of time (I find coding ANYTHING for ansible takes
3-5 times longer than it would to write a shell script, debugging to
be painful, and documentation to be barely adequate).

My concern with Xavier's work so far is that it increased the number
of files and duplication of data, and that I might be the only one
qualified to review it, and thus would have to wrap my head around it
or accept it as-is.

We DO NOT have the luxury of time, to have this be a multi-week
incremental crawl towards a solution.  In my mind, something that
works today is good enough, and is (at least) superior to written
documentation, because future people will at least know that it
worked, and was used at some point in time.

The one thing I would view as least desireable is any need for human
intervention in the install/configuration process.

pgulley:
Hi!
Many of the details of these changes are beyond my understanding, but I'd love to prioritize moving quickly towards a functional solution, rather than iterating towards the perfect ansible recipe here. I imagine we'll run this process possibly a small handful of times between now and when we properly deploy the new index, but ideally we won't revisit this process until far enough down the line that our requirements will have likely changed.

philbudne:
Q: Why is vault file needed?  If for sudo/become password there is another solution

philbudne:
OR if the vault is being used for snapshot repository keys:
1. Isn't this done by the elastic-conf script?
2. How will that work for staging (where ES is run from a stock docker image)?

And if the ansible configuration is creating index templates: what does that mean for dev/staging (ES run via docker)?

In the interests of KIS(S) and DRY: I'd like to suggest that not everything that _can_ be done by ansible should be.

philbudne:
AND if private config is needed, just put the unencrypted yaml file in the private config repo so it's all in one place?

PR #382, opened 2025-05-19
Chore/es ansible config

thepsalmist:
This PR adds Elasticsearch configuration using Ansible
1. Bootstrap initial index from `initial-index.json` file
2. Create index template and ILM policy
3. Modify ILM policy t rollover based on `max_primary_shard_size` only


conf/ansible/inventories/local/host_vars/localhost.yml
```diff
@@ -23,7 +23,11 @@ es_version: "8.17.3"
...
   network.publish_host: "_eth0_"
 
+mc_es_ilm_max_shard_size: "1gb"
+mc_es_replicas: 1
 mc_es_seed_hosts:
   - "es-node-01"
   - "es-node-02"
   - "es-node-03"
+
+mc_es_shards: 2
```
kilemensi, line 33:
```suggestion
mc_es_replicas: 1
mc_es_seed_hosts:
  - "es-node-01"
  - "es-node-02"
  - "es-node-03"
mc_es_shards: 2
```

They are all `mc_es_xxxx` configs.

conf/ansible/playbooks/es-configure.yml
```diff
@@ -0,0 +1,64 @@
...
+        headers:
+          Content-Type: "application/json"
+      register: es_info
+      until: es_info.status == 200
+      retries: 6
+      delay: 10
+      when: not is_staging_or_local | bool
+      tags: always
+
+    - name: Verify cluster health for all nodes (staging/local)
```
kilemensi, line 31:
```suggestion
    - name: Ensure Elasticsearch is accessible (staging or local)
```

Since the task is doing the same thing, why not use similar name? Or why are we verifying "cluster" in PROD but individual nodes in staging/local?

conf/ansible/scripts/es-configure.sh
```diff
@@ -0,0 +1,19 @@
...
+run_base "$@"
+
+playbook="../playbooks/es-configure.yml"
+
+echo "Running playbook with:"
+echo "  Inventory: $inventory"
+echo "  User: $user"
+[ -n "$env" ] && echo "  Environment: $env"
+[ -n "$extra_args" ] && echo "  Extra args: $extra_args"
+echo ""
```
kilemensi, line 17: I think these same exact 6 lines are repeated across all scripts... shouldn't they be part of base or a reusable function?

kilemensi (approved):
Code LGTM!

conf/ansible/scripts/es-configure.sh
thepsalmist: Sure, should consider this when doing a refactor

pgulley:
Hey! Looks good! Now that the configuration is centralized, is there support for dev/staging environments with this stack, or is additional work necessary to actually make those deployments possible? As an ansible layperson I'd appreciate some more documentation about the processes involved!

PR #387, opened 2025-05-21
Add a step to configure available vm in docker

pgulley:
Noticed this issue was causing failed deployment on a new host without other docker configuration, this is an easy fix.


kilemensi (approved):
🚀

---

The question is do we want to do this in `es-install-docker.sh` or not. Since this script doesn't touch any system settings (just deploys the cluster), it feels like this should be done by whatever does the actual docker installation on the system.

Issue #388, opened 2025-05-22
DevOps deployment options for NewsScribe cluster and NS Story Indexer


pgulley:
Since this is a critical piece of infrastructure, with multiple modes of deployment, we need a relatively robust set of deployment options for developers, and protections built around production. The existing story-indexer deployment scripts (managed via `docker/deploy.sh`) builds images from source, bound to a specific checked-out commit. Each deployment type ("hist", "archive", etc) has its own deployment namespace (with dev/staging/prod, eg "hist-staging-indexer") and dev instances are additionally disambiguated by the USER who is deploying them (ie, 'pgulley-indexer'). Currently, dev and staging stacks also deploy a containerized ES index for testing purposes. We need our deployment options  for the news-scribe cluster Elasticsearch index to have parity with these options, and since the production deployment itself is done with ansible, I'd love to have some documentation that we've thought through the corner cases that might come up for each style of deployment. 

Right now the ES Ansible setup has 'local/staging/dev'- so the missing pieces are the development deployment namespaces for USER and for REALM, which means naming the deployed dev instance accordingly, and providing port-offsets. Are there other pieces we need to lay out to make that work? 

Given the risk involved with this piece of our infrastructure, its important that we have a devops solution in place before moving the NewsScribe cluster all the way into production- so I'd like to see this sorted before we start indexing to the NewsScribe cluster.

thepsalmist:
### Multi-Cluster Setup 
We should be able to successfuly configure multiple clusters on both development and staging environments to mirror the functionality of `deploy.sh`. This implementation is currently being developed in **draft PR #[PR_NUMBER]**.

We can access users when running playbooks, therefore we should be able to have the `USER-stack` setup similar to the existing dev setup by `deploy.sh`.

## Integration Options Elasticsearch Ansible & Story indexer

We have identified two viable approaches for integrating Ansible deployment and configuration with the story indexer services:

### Option 1: Sequential Execution
- Execute the Ansible script to deploy and configure Elasticsearch
- On successful completion, extract the `ELASTICSEARCH_HOSTS` variable
- Pass this variable to the `deploy.sh` script for further processing

### Option 2: Automated Integration
- Invoke the Elasticsearch installation and configuration script directly from within `deploy.sh`
- Capture the output containing `ELASTICSEARCH_HOSTS`
- Automatically set this variable within the `deploy.sh` script execution flow

**Current Blocker on Automated Integration: Privilege Escalation

The Challenge on option 2:
[...]

PR #389, opened 2025-05-23
Setup for multiple cluster deployment, docker using Ansible

thepsalmist:

PR #384, opened 2025-05-20
Chore/es snapshots config using  ansible

m453h:
This PR adds the functionality to configure Elasticsearch for snapshot lifecycle management (SLM) and repository creation. It includes changes to Ansible playbooks, tasks, and templates to automate these processes.

Addresses #376

PR #392, opened 2025-05-26
Fix #391,388 -- update deploy.sh, elastic_conf for Newsscribe cluster

### LINKED ISSUE #391: #388 PRs implementation process
This issue outlines the PR implementation approach based on the discussions on #388

1. Limit Ansible to only PROD cluster installation (ES/JVM)
2. Revert deployment of dev/staging to `deploy.sh`, this includes moving the templates not directly required for ES install in 1, to `elasticsearch/templates`
3. End to end test of story-indexer stack with the Neswcribe dev/staging ES cluster
4. Cleanup non-required Ansible configurations (local/staging)

thepsalmist:
Implementation for #391 #388

### DIFF
--- a/conf/elasticsearch/templates/bi_weekly_slm_policy.json
+++ b/conf/elasticsearch/templates/bi_weekly_slm_policy.json
@@ -1,0 +1,14 @@
+{
+    "schedule": "0 30 1 1,15 * ?",
+    "name": "<snapshot-{now/d}>",
+    "repository": "$repository",
+    "config": {
+        "indices": ["mc_search-*"],
+        "ignore_unavailable": true,
+        "include_global_state": true
+    },
+    "retention": {
+        "min_count": 5,
+        "max_count": 10
+    }
+}
--- a/conf/elasticsearch/templates/create_ilm_policy.json
+++ b/conf/elasticsearch/templates/create_ilm_policy.json
@@ -1,0 +1,18 @@
+{
+    "name": "mediacloud-lifecycle-policy",
+    "policy": {
+      "phases": {
+        "hot": {
+          "actions": {
+            "rollover": {
+              "max_primary_shard_size": "$max_primary_shard_size"
+            }
+          }
+        }
+      },
+      "_meta": {
+        "description": "ILM policy using the hot phase with a rollover based on max_primary_shard_size",
+        "managed": true
+      }
+    }
+}
--- a/conf/elasticsearch/templates/create_initial_index.json
+++ b/conf/elasticsearch/templates/create_initial_index.json
@@ -1,0 +1,8 @@
+{
+    "name": "mc_search-000001",
+    "aliases": {
+        "mc_search": {
+        "is_write_index": true
+        }
+    }
+}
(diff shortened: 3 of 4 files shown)


kilemensi (approved):
👍🏽 

---

Of course, we shouldn't merge this to `main` until we've discussed it with MC folks when they're back online. If we have more than 1 PR to do, lets create a "staging" branch e.g. `non-ansible`, etc., where we can merge all relevant PRs first.

kilemensi:
Is it correct to assume this PR will make #384, #387, and #389 unnecessary moving forward?

thepsalmist:
> Is it correct to assume this PR will make #384, #387, and #389 unnecessary moving forward?

Yes sure

pgulley (approved):

philbudne (commented):
Since the variables ELASTICSEARCH_MAX_WAIT_TIME and ELASTICSEARCH_WAIT_INTERVAL apply only to elastic-conf.py, I think they should have names that are less generic, to avoid confusion.

Maybe ELASTICSEARCH_CONF_XXX? or ELASTIC_CONF_XXX???

Is there an explanation somewhere of why this change is needed? Does it take longer for the larger cluster to converge on startup??

thepsalmist:
> Since the variables ELASTICSEARCH_MAX_WAIT_TIME and ELASTICSEARCH_WAIT_INTERVAL apply only to elastic-conf.py, I think they should have names that are less generic, to avoid confusion.
> 
> Maybe ELASTICSEARCH_CONF_XXX? or ELASTIC_CONF_XXX???
> 
> Is there an explanation somewhere of why this change is needed? Does it take longer for the larger cluster to converge on startup??

docker stack deploy does not support `depends_on` [see here](https://github.com/moby/moby/issues/31333), and therefore we cannot guarantee that the `elastic-conf` service will apply configurations only when cluster is ready & healthy. We therefore implement our own polling to guarantee this

philbudne:
I asked:
> Is there an explanation somewhere of why this change is needed?
> Does it take longer for the larger cluster to converge on startup??

> docker stack deploy does not support `depends_on` [see
> here](https://github.com/moby/moby/issues/31333), and therefore we
> cannot guarantee that the `elastic-conf` service will apply
> configurations only when cluster is ready & healthy. We therefore
> implement our own polling to guarantee this

You didn't answer either of my questions, although I confess it might
have been clearer if I had asked "why is this change needed NOW?"

I'm not asking why the change MIGHT be helpful, I'm asking the
question I ask myself when I do a "git diff" before opening a P.R.:

I this NECESSARY (within scope of the issue)?  And if not, is now a
good time to be making "it would be nice if...."  changes?  For me,
the answer to the second question is no.

I could also have asked if running "make update" was necessary.

I don't think this is the time to be making extraneous changes, it
only prolongs a process which has already taken far longer than anyone
would have wished, and while it's tempting to hit the "approve" button
just to make the issue go away, I feel I'd be stopping short of my
ethical and educational responsibilities by doing that.

While I didn't necessarily agree, I seem to recall @kilemensi saying
about this (or something similar) to let Docker take care of
restarting the script if ES isn't ready.  I PERSONALLY have other
issues about the ES startup sequence (again because there is no way to
delay startup of a service until another service is ready), but I'm
not advocating making those changes now.

PR #409, opened 2025-07-10
Have hist-queuer discard rows without downloads_id

philbudne:
Currently passing everything thru to hist-fetcher, which quarantines them.

Did not change hist-fetcher; It's now reasonable to quarantine stories which should never have gotten there!

### DIFF
--- a/indexer/workers/hist-queuer.py
+++ b/indexer/workers/hist-queuer.py
@@ -85,7 +85,7 @@ class HistQueuer(Queuer):
                 url = NEED_CANONICAL_URL
 
             dlid = row.get("downloads_id")  # REQUIRED
-            if not dlid or not isinstance(dlid, str) or not dlid.isdigit():
+            if not dlid or not dlid.isdigit():
                 self.incr_stories("bad-dlid", url)
                 continue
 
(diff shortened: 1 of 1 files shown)


indexer/workers/hist-queuer.py
```diff
@@ -84,8 +84,10 @@ def process_file(self, fname: str, fobj: BinaryIO) -> None:
             else:
                 url = NEED_CANONICAL_URL
 
-            dlid = row.get("downloads_id")
-            # let hist-fetcher quarantine if bad
+            dlid = row.get("downloads_id")  # REQUIRED
+            if not dlid or not isinstance(dlid, str) or not dlid.isdigit():
```
kilemensi, line 88: Isn't everything returned by any of [`csv.reader`](https://docs.python.org/3/library/csv.html#csv.reader)s a `str` by default (if present)?

kilemensi (approved):

indexer/workers/hist-queuer.py
philbudne: Will fix.

I'll plead an overabundance of caution.  But mypy doesn't complain without the isinstance.

I can easily imagine there's a mypy setting to complain about constant sub-expressions, but I can also imagine it could easily find other such places!

pgulley (approved):

Issue #411, opened 2025-07-16
"Sump-pump" for filtering out problem content stuck in the refetch queues.


pgulley:
It's possible for the fetching processes to gum up if the fetcher queues get full of error-throwing stories. In this case, we'd like to be able to 'flush' the queues- but don't want to throw the baby out with the bath water. We need a standard process for filtering only stories that match some specific criteria from these queues, leaving the rest (EX: today the queues are full of a bunch of stories from one source that we know to be broken- we'd like to remove all and only stories that match that source, leaving whatever else might be swimming in the queues) 

@philbudne mentioned filtering out warcs, presumably dumped from the queue- 
Another approach might be to introduce a 'filter' story-indexer process that just passes stories through unchanged unless they meet a filter criteria- and have it read from the queues. 
Should be low-bandwidth to deploy, so whatever is easier.

kilemensi:
Not sure if it'll work in this case but RabbitMQ has [shovel plugin](https://www.rabbitmq.com/docs/shovel).

PR #415, opened 2025-07-21
qutil.py: honor x-mc- headers in reload_archives

philbudne:
This honors previous error counts, so the story won't be retried again from square zero.  Adds StoryArchiveReader.read_stories2() that yields both Story and metadata JSON.

fixes https://github.com/mediacloud/story-indexer/issues/411

### DIFF
--- a/indexer/storyapp.py
+++ b/indexer/storyapp.py
@@ -227,7 +227,7 @@ class RabbitMQStorySender(StorySender):
         exchange: Optional[str] = None,
         routing_key: str = DEFAULT_ROUTING_KEY,
         expiration_ms: Optional[int] = None,
-        headers: HeaderDict = {},
+        headers: HeaderDict = {},  # for qutil load_archives command
     ) -> None:
         if self._batch_size <= 1:
             # avoid overhead if not batching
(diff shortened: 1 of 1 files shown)


indexer/story_archive_writer.py
```diff
@@ -411,8 +414,10 @@ class StoryArchiveReader:
     def __init__(self, fileobj: BinaryIO):
         self.iterator = ArchiveIterator(fileobj)
 
-    def read_stories(self) -> Iterator[BaseStory]:
-        # read WARC file:
+    def read_stories2(self) -> Iterator[tuple[BaseStory, MetaDataDict]]:
```
kilemensi, line 417: Don't have full context so I could be wide of the mark but why not:

```python
    def read_stories(self, with_metadata=False): # default to False for backward compatibility
```
or
```python
    def read_stories_with_metadata(self):
```
The name says why this one is different from the existing `read_stories`

indexer/storyapp.py
```diff
@@ -245,9 +251,14 @@ def _send(
         exchange: Optional[str],
         routing_key: str,
         expiration_ms: Optional[int],
+        headers: Optional[HeaderDict] = None,
```
kilemensi, line 254: Looking at definition of `_BatchItem`, `send_story`, etc., looks like `headers` is not `Optional`?

kilemensi (approved):

indexer/storyapp.py
philbudne: I changed RabbitMQStorySender.send_story `headers` argument to default to empty dict, and is no longer "Optional" (possibly None).  @kilemensi does this address your issue/question (took me a while to realize that the problem could be THAT simple)!

kilemensi: Yep! it does @philbudne

PR #444, opened 2026-05-20
indexer/workers/pipeview/api.py: Pydandic model for pipeview api params

philbudne:
I'm always accessing the pipeview API with curl from the command line, and it's helpful to have it yell at me when I make a typo in a parameter name!


rahulbot (approved):
Look useful

